PR #18, opened 2019-09-05
give some guidelines about github usage

grahamlopez:


contributing/index.rst
```diff
@@ -75,3 +75,19 @@ Submit suggestions
...
+GitHub Guidelines
+===================
+
+Here are some guidelines and common practices that we use in this project.
+
+- When you want to work on an issue, assign it to yourself if no one is assigned
+  yet. If there is somebody assigned, check in with that person about
+  collaborating.
+- Reference the issue(s) that your PR addresses with GitHub's '#' notation.
+- Use "WIP" in your PR title to indicate that it should not be merged yet.
```
grahamlopez: This line is awkward

grahamlopez:
:man_shrugging: I'm indifferent. It seems redundant to me, since it doesn't change the notifications that we get by default. Put another way: What issue /wouldn't/ need attention from this team?

jack-morrison:
I imagine there may be situations where it could be helpful to bring in several of us from UA for discussion and the GH team is the builtin way to do this.

Issue #14, opened 2019-09-05
re-sync Rhea user guide


grahamlopez:
The wordpress-based Rhea User Guide at https://www.olcf.ornl.gov/for-users/system-user-guides/rhea/ is frozen as of 06 September 2019.

This means that the content in these new pages needs to be re-synced. Here is the procedure:
## howto
1. put your name by a section below to claim it
2. update the content of that section:
    - make sure that the topics covered by the previous wordpress page are covered here
    - check that all hyperlinks (both external and internal work). Actually click on them; don't just read them
    - check visual elements: images scale correctly, tables are shaped correctly, etc.
    - fix any other problems you see, or at least submit a separate issue for it to remind us to fix it in the future
    - think about what could make the section better, and submit issue(s) with suggestions
3. Be sure to mention this issue via something like "Partially addresses #14" in your PR description.
4. check off your section below once your PR addressing it has been merged.
5. return to "1" above

## todo
- [x] System Overview @grahamlopez 
- [x] Connecting* @isubsmith 
- [x] Data Storage & Transfers* @gmarkomanolis 
- [ ] Software* @isubsmith 
- [x] Shell & Programming Environments @grahamlopez 
- [x] Compiling @gmarkomanolis 
- [x] Running Jobs @isubsmith 
- [x] Analysis Tools @gmarkomanolis 

*these sections should just refer back to the general pages on the topic, possibly with exceptions mentioned for Rhea

porgull:
@grahamlopez PR ready for shell/programming env
@isubsmith PR ready for Running Jobs section
@gmarkomanolis PR ready for Analysis/tools section

isubsmith:
See PR #55  for "Connecting" changes.

grahamlopez:
software section is a non-blocking work in progress. these are ready to go live.

PR #25, opened 2019-09-06
Rhea compiling

gmarkomanolis:
Rhea compiling section was checked. Fixed links and added internal references to make easier the navigation.


grahamlopez:
partially addresses #14

PR #42, opened 2019-09-09
Fix rhea system overview

porgull:
partially addresses #14


systems/includes/rhea-compute-nodes.txt
```diff
@@ -14,16 +14,16 @@ Rhea contains 521 compute nodes separated into two partitions:
```
grahamlopez: I don't really see any reason to use a separate file for this section. Let's just go ahead and inline this in the right spot in the user guide.

systems/includes/rhea-compute-nodes.txt
```diff
@@ -35,7 +35,7 @@ With Hyper-Threading enabled, these nodes have 56 logical cores that can execute
 
 Please see the :ref:`batch-queues-on-rhea` section to learn about the queuing
 policies for these two partitions. Both compute partitions are accessible
-through the same batch queue from Rhea’s login nodes.
+through the same batch queue from Rhea’s `login nodes <#overview>`__.
```
grahamlopez: can we make this a cross-ref?

porgull: yep, committed

systems/includes/rhea-compute-nodes.txt
porgull: it is now in-line

porgull:
@grahamlopez the links to filesystems don't work in this PR, but those are tentative, so should that be handled in a separate issue?

grahamlopez:
@porgull yes, we will have to fix those later.

PR #44, opened 2019-09-09
Fix rhea shell and programming environment section

porgull:
partially addresses #14


porgull:
@grahamlopez added commit to move order

grahamlopez:
thank you!

PR #49, opened 2019-09-10
Remove 'content block' from vampir section

isubsmith:
Very minor change, just wanted to go through the process once with something minor.


grahamlopez:
looks like this is already done as part of #36?

isubsmith:
> looks like this is already done as part of #36?

Hmm. It wasn't when I checked out this morning?

PR #55, opened 2019-09-11
Add connecting section, pull from systems guides (and point back to main connecting section)

### LINKED ISSUE #9: "Connecting" - initial conversion or rewrite
This is a top-level section, and also needs to be linked to in the relevant places in the system user guides, and elsewhere as appropriate.

isubsmith:
Pulled all the "connecting" text out of system user guides (minus "summit is at ssh username@summit.olcf.ornl.gov") and created a new top level connecting section. 

closes #9


grahamlopez:
@jack-morrison @isubsmith I think I've got it. See what you think.

getting_started/getting_started.rst
```diff
@@ -209,6 +209,8 @@ Home can be accessed through ssh utilities:
 
       ssh userid@home.ccs.ornl.gov
 
+For more information on connecting to OLCF resources, see the <.connecting>`__ section.
```
grahamlopez: I just noticed that we have two duplicated sections "Systems Available to All Projects" and "Connecting for the first time" in both the "Getting Started" and "Connecting" sections. It would be good to reconcile these somehow so we don't have duplicated content that can drift out of sync. Probably one should refer back to the other?

grahamlopez:
@jack-morrison @isubsmith I factored out the duplicated sections from "Getting Started." See what you think. I'm good for this one to go when you are.

Issue #13, opened 2019-09-05
re-sync Summit user guide


grahamlopez:
The wordpress-based Summit User Guide at https://www.olcf.ornl.gov/for-users/system-user-guides/summit/summit-user-guide/#system-overview is frozen as of 06 September 2019.

This means that the content in these new pages needs to be re-synced. Here is the procedure:
## howto
1. put your name by a section below to claim it
2. update the content of that section:
    - make sure that the topics covered by the previous wordpress page are covered here
    - check that all hyperlinks (both external and internal work). Actually click on them; don't just read them
    - check visual elements: images scale correctly, tables are shaped correctly, etc.
    - fix any other problems you see, or at least submit a separate issue for it to remind us to fix it in the future
    - think about what could make the section better, and submit issue(s) with suggestions
3. Be sure to mention this issue via something like "Partially addresses #13" in your PR description.
4. check off your section below once your PR addressing it has been merged.
5. return to "1" above

## todo
- [x] Summit Documentation Resources @grahamlopez 
- [x] System Overview @jack-morrison 
- [x] NVIDIA V100 GPUs @tom-papatheodore 
- [x] Connecting* @isubsmith 
- [x] Data Storage & Transfers* @gmarkomanolis 
- [ ] Software* @isubsmith 
- [x] Shell & Programming Environments @isubsmith 
- [x] Compiling @grahamlopez 
- [x] Running Jobs @gmarkomanolis 
- [x] Debugging @grahamlopez  
- [x] Optimizing and Profiling @isubsmith 
- [x] Known Issues @grahamlopez 
- [x] Training System (Ascent) @tom-papatheodore 
- [x] Preparing for Frontier @tom-papatheodore 

*these sections should just refer back to the general pages on the topic, possibly with exceptions mentioned for Rhea

porgull:
@grahamlopez  @jack-morrison  @tom-papatheodore 
Created PRs for the first three sections; it's just fixing newlines/whitespace that existed on olcf.ornl.gov that didn't here. No other changes necessary.

porgull:
@isubsmith @grahamlopez @gmarkomanolis 

Created PRs for shell, compiling, and running jobs sections.

porgull:
@tom-papatheodore @isubsmith  @grahamlopez

Added rest of sections that aren't being moved.

isubsmith:
See PR #55  for "Connecting" changes.

grahamlopez:
[...]

PR #56, opened 2019-09-11
Update compiler section

isubsmith:
Fairly minor - add the proper defaults (and a date for when they were defaults) for compiler versions. Addressees #13  compiler section.


isubsmith:
Damn it, has my unmerged "connecting" work too. Will generate a clean(er) patch.

PR #57, opened 2019-09-11
Update summit compiling section

isubsmith:
Addresses compilers section from #13 

Update default versions (and date stamp the update) for compilers table. Minor wording/typo changes.


grahamlopez:
I pushed some changes to use more code fences for command options and to try to shrink some table widths just a little bit

grahamlopez:
@isubsmith if you are cool with my tweaks, I'm okay to merge

isubsmith:
Yup, I'm good with those changes.

systems/summit_user_guide.rst
```diff
@@ -1182,17 +1182,17 @@ C compilation
 
 +--------------+------------------+----------------+------------------+------------------+------------------+---------------------------+--------------------+
 | **Vendor**   | **Module**       | **Compiler**   | **Default**      | **Enable C99**   | **Enable C11**   | **Default signed char**   | **Define macro**   |
-|              |                  |                | - (10 Sept 2019) |                  |                  |                           |                    |
+|              |                  |                |                  |                  |                  |                           |                    |
```
grahamlopez: I had left those in as I thought the point was to tell users something about the timeliness/validity of the information.

Thinking about it further, I think this is an un-winnable game. Maybe we shouldn't list defaults at all?

jack-morrison: I think this should be rolled into conversation around #5. @isubsmith and I are meeting tomorrow morning to discuss. As you mention, it may be best to remove the default versions here.

This PR removed them with the idea that all information in the user guides should (ideally) be current.

grahamlopez:
> e1e4eab removed them with the idea that all information in the user guides should (ideally) be current.

I can get behind that, but it might be nigh impossible in practice. Hence just not saying what the default is and taking the @mpbelhorn approach of "if you want the most current information about specifics, query it directly from the system."

systems/summit_user_guide.rst
grahamlopez: I can get behind that, but it might be nigh impossible in practice. Hence just not saying what the default is and taking the @mpbelhorn approach of "if you want the most current information about specifics, query it directly from the system."

jack-morrison: Agreed. That'll always be the best answer imo.

isubsmith: I'm fine with that too (dropping version info). The info was so far out of date I felt like I needed to add the date for when it was current. It probably makes more sense to leave the versions available and default off entirely, and then provide a Note: or something with how to find the latest and the default version if you are logged on.

PR #64, opened 2019-09-13
Fixes #63

### LINKED ISSUE #63: Remove `Indices and tables` from index.rst
Looks like this section of the main `index.rst` is from the initial [bare bones reST setup](https://github.com/olcf/olcf-user-docs/commit/32359e98dfdd011d3e9d935e1ffd71287e8e0ffa#diff-db4b047b30ffef7855b12b527860ac2c), and can probably be removed.

jack-morrison:
removes unused stuff from vanilla RTD

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -59,10 +59,3 @@ How to contribute to this documentation
    :maxdepth: 2
 
    contributing/index
-
-Indices and tables
-==================
-
-* :ref:`genindex`
-* :ref:`modindex`
-* :ref:`search`


index.rst
```diff
@@ -59,10 +59,3 @@ How to contribute to this documentation
    :maxdepth: 2
 
    contributing/index
-
-Indices and tables
```
grahamlopez: Yeah, I had left these in because I didn't know if they would potentially be useful. I haven't gotten to the point of educating myself enough to know if we could make use of any of these. Thoughts?

jack-morrison: I'm not really sure what to do with them either. I'd be fine cutting them out for now.

PR #80, opened 2019-09-24
Updated wording on login node cgroups

### LINKED ISSUE #79: Alter login node cgroup wording

chrisfuson:
Fixes #79

### DIFF
--- a/systems/summit_user_guide.rst
+++ b/systems/summit_user_guide.rst
@@ -1452,13 +1452,26 @@ Othewise, you will not have any compute nodes allocated and your parallel
 job will run on the login node. If this happens, your job will interfere with
 (and be interfered with by) other users' login node tasks.
 
-Because the login nodes are resources shared by all Summit users, we utilize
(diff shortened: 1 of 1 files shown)


chrisfuson:
@jack-morrison , incase it helps, I updated the wording on the production site to the following:

Per-User Login Node Resource Limits
Because the login nodes are resources shared by all Summit users, we utilize cgroups to help better ensure resource availability for all users of the shared nodes. By default each user is limited to 8 hardware-threads, 8GB of memory, and 1 GPU. Please note that limits are set per user and not individual login sessions. All user processes on a node are contained within a single cgroup and share the cgroup’s limits.

In addition to the default settings, once one of a user’s processes on a node reaches 4-hours of CPU-time, all of the user’s processes on the node will be limited to .5 hardware-thread. Once the CPU time reduction has been triggered, you can reset the node’s cgroup limits to default by first killing the long running process and then by starting a new login session to the node. Note, you do not have to log out of the session to reset the node’s defaults once the long running process has been killed, but you will need to start a new session on the same node.

PR #83, opened 2019-09-25
Minor updates

mpbelhorn:


grahamlopez:
Looks fine to me. If we agree we like the line numbers in the code listings, we should make an issue to make this consistent throughout (summit guide, etc.). Whomever ends up merging, please do this.

mpbelhorn:
@grahamlopez Those were the only code blocks that I saw which referenced line numbers in explanation text. I think that generally the line numbers are not necessary and should be omitted. But when they are referred to in text, they should be present and not part of the code snippet since... well... they're not code. Also, if sphinx generates them, users can easily copy and paste the actual code.

grahamlopez:
@mpbelhorn ah cool. makes sense to me!

PR #100, opened 2019-09-30
Update wording of Data Retention Policy in Data Guide

jack-morrison:


data/policies.rst
```diff
@@ -53,11 +53,10 @@ these systems regularly. See :ref:`hpss`. for information about using the HSI
...
-that case, the data will be retained according to the purge policy. After 90
-days, the data will not be available but not purged for another 60 days, where
-the data will be removed if not requested otherwise.
+Data Retention
+==============
+
+By default, the OLCF does not guarantee lifetime data retention on any OLCF
+resources. Following a user account deactivation or project end, user and
+project data in non-purged areas will be retained for 90 days. After this
+timeframe, the OLCF retains the right to delete data.
```
grahamlopez: I know this was discussed quite a lot already internally, but should add a sentence at the very end: "Data in purged areas remains subject to the normal purge policies."

jack-morrison: I'm on board with that. Updated.

PR #110, opened 2019-10-01
Add some examples for scp and rsync

gmarkomanolis:
Adding data transfer examples with some command line tools


grahamlopez:
I made some edits to this one, so would like to get a different set of eyes on it for final merge.

PR #140, opened 2019-11-01
Profiling

gmarkomanolis:
Starting with TAU tool


grahamlopez:
we should also rebase before merging

PR #152, opened 2019-11-21
Jinja Templated RST

frobnitzem:
This PR adds jinja template processing to allow inserting code sections into rst files within ``{{ }}`` quotes.  The immediate application is to auto-generate most of the tables with filesystem properties in the documentation -- so all their information can be updated in one single (yaml) data source.

It also requires the yaml package for python.  Sphinx already uses jinja, so there is no extra jinja dependency.

### DIFF
--- a/accounts/olcf_policy_guide.rst
+++ b/accounts/olcf_policy_guide.rst
@@ -223,23 +223,7 @@ categories: those intended for user data and those intended for project
 data. Within each of the two categories, we provide different sub-areas,
 each with an intended purpose:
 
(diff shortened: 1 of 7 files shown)


grahamlopez:
Hi @frobnitzem. We have discussed this proposed change internally, and for now we would like to keep things simpler (even at the cost of some manual effort) for ease of maintenance and a lower barrier to entry for contributors who aren't as comfortable with RST, Sphinx, etc.

I want to emphasize that we appreciate your contributions and ideas; please do continue to make suggestions as you see things that you think can be improved.

Issue #154, opened 2019-11-25
Update NiceDCV info in Rhea User Guide to use Slurm


warenaud:

jack-morrison:
Closed with #155

PR #155, opened 2019-11-25
Nicedcv to slurm

warenaud:
Removed the step showing how to get an interactive job with Torque/Moab and replaced it with the salloc equivalent.


warenaud:
Addresses #154

Issue #156, opened 2019-11-27
Change VisIt documentation for Rhea to reflect Slurm, other minor Slurm-related changes


isubsmith:

jack-morrison:
Closed by #157

PR #157, opened 2019-11-27
Fix up some Rhea moab->slurm stuff, consistentify user{*} to username

isubsmith:
Changes for issue #156


jack-morrison:
Closes #156

PR #161, opened 2019-12-04
Moves FAULT_TOLERANCE Issue to Resolved

jack-morrison:
This is a partial review of Summit's Known Issues section.

The main change here is a move of the `FAULT_TOLERANCE` issue to resolved. Previously, setting `FAULT_TOLERANCE=1` in `~/.jsm.conf` would cause job launch to fail. This is reported fixed by IBM, and I have confirmed with a small job that launch works when this is set in the current software stack.

### DIFF
--- a/systems/summit_user_guide.rst
+++ b/systems/summit_user_guide.rst
@@ -3125,7 +3125,7 @@ please see the `Vampir Software Page <https://www.olcf.ornl.gov/software_package
 Known Issues
 ============
 
-Last Updated: 09 September 2019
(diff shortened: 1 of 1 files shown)


jack-morrison:
Something I discovered when making these changes... The "Last Updated" timestamp implies that the entire "Known Issues" section was reviewed. This isn't necessarily the case, as in this PR. I only moved one of the issues. 

Is this a problem? If so, I'd be open to discussing other formats of timestamping issues. 

Attn: @olcf/ua-support

PR #173, opened 2019-12-06
Add description of CUDA hook issue and workarounds

### LINKED ISSUE #78: known issue: cuda hooks / pami
When running simple gpu codes with jsrun that do not also have MPI support, sometimes one might run into a warning such as:

`CUDA Hook Library: Failed to find symbol mem_find_dreg_entries, ./a.out: undefined symbol: __PAMI_Invalidate_region`

This can be solved in a few ways:
- use `jsrun -E LD_PRELOAD=/opt/ibm/spectrum_mpi/lib/pami_451/libpami.so ...`
- use `jsrun --smpiargs="off" ...`
- use `jsrun --smpiargs="-disable_gpu_hooks ...`

Some more discussion can be found at https://github.com/kokkos/kokkos/issues/1985; reports there say it has been reported to IBM.

maxpkatz:
This PR describes the common issue that occurs when a serial application is launched with jsrun and gets a CUDA hook error. This can also happen if you call into CUDA before MPI_Init(). Two workarounds are proposed. Note that while I'm submitting this to Open Issues, this is really a design choice by IBM that has been carefully considered and has reasons for being designed the way it is, so it is not something that they are intending to change in the near future as far as I know.

Closes #78

### DIFF
--- a/systems/summit_user_guide.rst
+++ b/systems/summit_user_guide.rst
@@ -3167,6 +3167,47 @@ Last Updated: 04 December 2019
 Open Issues
 -----------
 
+CUDA hook error when program uses CUDA without first calling MPI_Init()
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Serial applications, that are not MPI enabled, often face the following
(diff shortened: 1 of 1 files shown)


grahamlopez:
thanks for the details here, @maxpkatz

PR #181, opened 2019-12-11
Score-P

gmarkomanolis:
Preparing the material for Score-P tutorial


jack-morrison:
Hey @gmarkomanolis -- Thanks for putting this together!

I started reviewing this today, and unfortunately it looks like MiniWeather has moved to a CMake build system after this MR was open as a draft. Would you mind updating the instrumentation example(s) to match how MiniWeather is built now? I started doing so in 81be937. 

I also noticed that the following two sections are only code snippets:
`Automatic Source Code Instrumentation > Prefix method`
`Automatic Source Code Instrumentation > Wrapper method`

Could you add a blurb about the difference between these two methods?

You'll see in 0e1eceb there's quite a bit of table re-formatting, which was just to bring things into alignment and emphasize values as in-lined code.

Thanks again! Happy to finally be getting this one published.

PR #186, opened 2019-12-19
Enable horizontal scrolling for tables at lower widths

### LINKED ISSUE #91: Tables not visible at low width
Body content is responsive for smaller screens / browser windows. Tables are _somewhat_ responsive, but a table's responsiveness is limited by the amount of content in its cells. When tables cannot be dynamically made smaller, the rightmost columns become inaccessible. Horizontal scrolling is an option, but is not working. 

Tested and confirmed a problem in:
- Chrome
- Firefox

Screen Shot 2019-09-27 at 1 54 49 PM

jack-morrison:
Fixes #91.

This change enables horizontal scrolling for tables of class `.wy-table-responsive` when the content would otherwise not be visible (small widths that aren't small _enough_ to toggle mobile media queries).

Note that to pick up this change, for review and in production, one may need to clear their browser's cached styles.

### DIFF
--- a/_static/css/theme_overrides.css
+++ b/_static/css/theme_overrides.css
@@ -16,6 +16,6 @@
    }
 
    .wy-table-responsive {
-      overflow: visible !important;
+      overflow: auto !important;
    }
 }


jack-morrison:
@grahamlopez would you mind reviewing this when you get a few? If you need a table to look at and test this change, there's an example in #91.

PR #192, opened 2020-01-23
Removes 'at the OLCF' from Software section title

### LINKED ISSUE #187: Change "Software at the OLCF" to "Software"
There's no need to include "at the OLCF" within OLCF documentation.

jack-morrison:
Fixes #187

### DIFF
--- a/software/index.rst
+++ b/software/index.rst
@@ -1,8 +1,8 @@
 .. _software-at-the-olcf:
 
-#####################
-Software at the OLCF
-#####################
+#########
+Software
+#########
 
 
 


jack-morrison:
Note that I left the label `_software-at-the-olcf` in `/software/index.rst`. 

Changing this to `_software` introduces a name collision with a label used in the Summit User Guide. Since it's much more likely that someone would have provided a direct link to the Summit User Guide > Software section, I left that one untouched. Don't want to be in the business of breaking links that exist in the wild.

grahamlopez:
I'm not disagreeing with your decision, but more of an FYI: changing internal links (that start with an underscore) shouldn't break any externally visible links. Those are generated from the actual section names.

grahamlopez:
Changing the underscore links would, however, break existing internal RST links.

jack-morrison:
```
$ grep -rnw './' -e '_software' --exclude "*build*"
.//software/index.rst:1:.. _software:
.//systems/summit_user_guide.rst:1064:.. _software:
```

There weren't any other internal RST links in this case. It's good to know that external links come from the section names.

PR #213, opened 2020-02-27
ibm-wml-ce: Changes for 1.7.0

bethune-bryant:
- Added the new package versions.
- Removed IBM DDL specific tuning parameters that aren't needed anymore.
- Added explicit information about loading specific module versions.

We should wait for @mpbelhorn to install the module, and I think we may be waiting for an announcement, since this version will be moving from Tensorflow 1 to 2 and may break people's scripts if they're just doing `module load ibm-wml-ce`

@bnemanich @nvcastet


jack-morrison:
Thanks for updating this, @bethune-bryant!

PR #222, opened 2020-03-06
ibm-wml-ce: Remove horovod build instructions and update versions.

bethune-bryant:
The horovod build instructions are no longer needed, and I updated the version numbers in some of the examples to be more up to date.


bethune-bryant:
FYI @jack-morrison, @bnemanich

PR #228, opened 2020-03-11
New versions of SW not Default SW

verolero86:
The second table just shows new versions deployed, not new defaults. This fixes that header.


jack-morrison:
@verolero86 sorry for the delay... I must've missed this one hitting my inbox. LGTM 🚀

PR #272, opened 2020-05-28
initial ecp ci document commit

proutrc:
This is to start documenting ECP CI at OLCF. 

The goal is to give a high level overview of what ECP CI is, how to get started at OLCF, and the details of how it is implemented at OLCF.


proutrc:
@tom-papatheodore @mpbelhorn @chrisfuson

tom-papatheodore:
That was quick work putting this together! Nice work! I pushed some minor changes to the existing docs and have a few comments/questions for your review:

- When would the "new project" option be used instead of the "mirroring" option? Will all teams setting up CI already have accounts/projects on Summit?

- I don't believe mirroring a moderate project to open automatically brings over the users too, since this would require all moderate users of a project to have XCAMS or UCAMS accounts already (which not everyone does). 

- Do users need to request an XCAMS account through the portal you linked to? I believe an XCAMS account will be created, along with an NCCS Open user in RATS, upon approval to the open-security project. If this is true, the "What Accounts Do I Need?" and/or "Applying For Project Access" sections might need to be modified.

- In the "Accessing the CI Infrastructure" section, it is mentioned that after the Open project has been set up, the requestor should submit a help ticket to get added to the ecpcitest group. Do we want the requestor to have to do this? E.g., if the accounts team knows the project is for CI work, they could open the ticket on the requestor's behalf to eliminate the extra step for the user. 

- Also in that section, it is stated that the users will have access to code.ornl.gov and login access to Ascent. Did we decide we wanted to allow CI users to directly log in to Ascent rather than limit them to access through only the CI processes?

PR #279, opened 2020-06-22
IBM WML-CE: Deprecate ddlrun.

### LINKED ISSUE #266: ddlrun should be replaced with jsrun for IBM WML documentation
The documentation on [IBM WML](https://docs.olcf.ornl.gov/software/analytics/ibm-wml-ce.html) uses ddlrun, but this is no longer best practice and we should convert these into equivalent jsrun commands.

bethune-bryant:
Deprecates the `ddlrun` tool and documents best practice for using `jsrun`.

Resolves #266

@bnemanich

### DIFF
--- a/software/analytics/ibm-wml-ce.rst
+++ b/software/analytics/ibm-wml-ce.rst
@@ -39,21 +39,21 @@ packages, and their dependencies:
 .. table::
     :widths: 20 40 40 35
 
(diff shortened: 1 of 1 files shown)


bethune-bryant:
@jack-morrison 
I just noticed that this never got merged. Is there something I need to do, or can we go ahead and merge it?

jack-morrison:
@bethune-bryant Good catch! I'm no longer with the OLCF, but I'm willing to bet @tom-papatheodore, @isubsmith, or someone else in the UA team would be happy to pick this up.

tom-papatheodore (approved):
LGTM.

PR #298, opened 2020-07-30
Migrating internal slate docs and updating docs for OpenShift v4

jeremyrogers:


jeremyrogers:
@kincl @proutrc ptal, might want to build it locally first to make sure all the images/links render properly. I think I got it all but a double check is always nice.

proutrc:
@jeremyrogers @kincl This looks really nice! I like the layout much better

PR #306, opened 2020-08-19
Adding GPU doc

kincl:


kincl:
Pretty bare bones at the moment, let me know if you think I should add something.

kincl:
> Looks like a minor typo - missing node count in "...Marble cluster has __ nodes with three NVIDIA..."

I was being intentionally ambiguous, do you think we should put the node count in documentation?

proutrc:
> > Looks like a minor typo - missing node count in "...Marble cluster has __ nodes with three NVIDIA..."
> 
> I was being intentionally ambiguous, do you think we should put the node count in documentation?

Ah, ok, I see. It is how I read the sentence then - being ambiguous does make sense.

PR #312, opened 2020-09-29
Updating Slate shared fs annotations

jeremyrogers:


services_and_applications/slate/access_olcf_resources/mount_fs.rst
```diff
@@ -10,8 +10,10 @@ to get the necessary configuration injected into the container at runtime.
   :header: "Cluster", "Annotation", "Value", "Mounts"
   :widths: 5, 8, 5, 25
 
-  "Marble", "ccs.ornl.gov/fs", "olcf", "/ccs/sw, /ccs/home, /ccs/proj, /gpfs/alpine"
+  "Marble", "ccs.ornl.gov/fs", "olcf", "/ccs/sw, /ccs/home, /ccs/home1, /ccs/home2, /ccs/sys, /ccs/proj, /gpfs/alpine"
```
kincl, line 13: No need to specify home[1,2] since that should be invisible to users

kincl (changes requested):

kincl:
@jeremyrogers

services_and_applications/slate/access_olcf_resources/mount_fs.rst
```diff
@@ -10,8 +10,10 @@ to get the necessary configuration injected into the container at runtime.
   :header: "Cluster", "Annotation", "Value", "Mounts"
   :widths: 5, 8, 5, 25
 
-  "Marble", "ccs.ornl.gov/fs", "olcf", "/ccs/sw, /ccs/home, /ccs/proj, /gpfs/alpine"
+  "Marble", "ccs.ornl.gov/fs", "olcf", "/ccs/sw, /ccs/home, /ccs/sys, /ccs/proj, /gpfs/alpine"
+  "Marble", "ccs.ornl.gov/fs", "olcf-nfs", "/ccs/sw, /ccs/home, /ccs/sys, /ccs/proj"
```
kincl, line 14: We mount /gpfs/alpine with olcf-nfs

services_and_applications/slate/access_olcf_resources/mount_fs.rst
```diff
@@ -10,8 +10,10 @@ to get the necessary configuration injected into the container at runtime.
...
   "Onyx", "ccs.ornl.gov/fs", "ccsopen", "/ccsopen/sw, /ccsopen/home, /ccsopen/proj, /gpfs/wolf"
+  "Onyx", "ccs.ornl.gov/fs", "open-nfs", "/ccsopen/sw, /ccsopen/home, /ccsopen/proj"
```
kincl, line 16: should this match "ccsopen" as "ccsopen-nfs"?

services_and_applications/slate/access_olcf_resources/mount_fs.rst
jeremyrogers: Cool, do we also mount (or do we plan to mount) /gpfs/wolf with open-nfs?

kincl: We don't right now, no plans to do so yet

services_and_applications/slate/access_olcf_resources/mount_fs.rst
jeremyrogers: Yes. I'll add that to the opa rules and then update here before merging. Do you think we should keep the "open-nfs" annotation as an undocumented option in case people are already using it?

kincl: good thought but I know no one is using it right now so we can drop it

jeremyrogers: Alright, updated the opa rules and changed it here

kincl (approved):

PR #324, opened 2020-11-03
making the menu screenshots visible

secondspass:
The screenshots for the Route section on the Slate docs were not visible so updating that.


michael-sandoval:
fixed in different PR

PR #335, opened 2020-12-14
change conda create env example to use rapids env

proutrc:


services_and_applications/jupyter/overview.rst
```diff
@@ -120,40 +120,48 @@ Conda environments need to be saved in a ``/ccs`` or ``/gpfs/alpine`` directory.
...
+   ``-c defaults rapids-blazing=0.17 python=3.7 cudatoolkit=10.2``
+   
+   * **NOTE**: The *conda create* command above assumes you are using **CUDA 10** JuptyerLab environment.
    
    * You need to use ``-p`` method if you want your environment to persist across Jupyter
      restarts. The path can be a location in ``/ccs`` or ``/gpfs/alpine`` that is writable
      by your user.
      
-#. Activate the environment ``source activate /path/to/tensorflowenv``.
+#. Activate the environment ``source activate /ccs/proj/<YOUR_PROJECT_ID>/<YOUR_UID>/rapids``.
```
secondspass, line 143: We should add a note here saying that the user should use `source activate` and not `conda activate` even though that is what is recommended at the end of the `conda create` process. Because conda activate requires extra setup and needs modification of their bashrc file.

secondspass (approved):

PR #345, opened 2021-01-21
Jupyter: ipympl hints

ax3l:
Add hints from support on how to install ipympl widgets.

- [ ] to-do: iron out last rendering/model view issues with support

### DIFF
--- a/services_and_applications/jupyter/overview.rst
+++ b/services_and_applications/jupyter/overview.rst
@@ -107,10 +107,13 @@ the Launcher and type ``conda list``. The libraries in the base environment shou
 cover most use cases. This base environment is provided to the default "Python 3" notebook, visible in the 
 JupyterLab "Launcher" page.
 
-From the Terminal, you can install additional libraries with a simple ``conda install`` to
+From the Jupyter Terminal, you can install additional libraries with a simple ``conda install`` to
 use in your current session. But these installed libraries won't persist across sessions
-if your server is restarted. 
+if your server is restarted.
 
+Since conda's dependency concretizer has sometimes problems to extend existing environments with new
+libraries, you can also use the `mamba <https://github.com/mamba-org/mamba>`_ package manager. Just
+install it with ``conda install -c conda-forge mamba`` and replace ``conda`` with ``mamba`` calls.
 
 
 Creating your own Conda environment and notebook entry
@@ -183,6 +186,23 @@ You may follow these general steps to create and install packages within an any 
  - *source activate /path/to/env*
  - *python -m ipykernel install --user --name <env-name> --display-name <env-name>*
  - *install packages with conda or pip in the activate environment*
+
+
+
+Example: Creating a Conda environment for Matplotlib Widgets
+============================================================
+
+We currently install Jupyterlab version 2, which needs additional treatment to get interactive extensions
+`such as matplotlib widgets installed <https://github.com/matplotlib/ipympl>`_.
+
+Here is an example how to install matplotlib:
+
+:: 
+   conda install -c conda-forge ipympl==0.5.8``
+   jupyter labextension install @jupyter-widgets/jupyterlab-manager jupyter-matplotlib@0.7.3
+
+Once the installation fishes, restart the kernel of your current notebook and refresh the page.
+You can now use the `%matplotlib ipympl` magic before importing matplotlib to render interactive plots.
 
 
 Manually stopping your JupyterLab session


services_and_applications/jupyter/overview.rst
```diff
@@ -185,6 +188,23 @@ You may follow these general steps to create and install packages within an any
...
+Example: Creating a Conda environment for Matplotlib Widgets
+============================================================
+
+We currently install Jupyterlab version 2, which needs additional treatment to get interactive extensions
+`such as matplotlib widgets installed <https://github.com/matplotlib/ipympl>`_.
+
+Here is an example how to install matplotlib:
+
+:: 
+   conda install -c conda-forge ipympl==0.5.8``
```
ax3l, line 201:
```suggestion
   conda install -c conda-forge ipympl==0.5.8
```

services_and_applications/jupyter/overview.rst
```diff
@@ -185,6 +188,23 @@ You may follow these general steps to create and install packages within an any
  - *install packages with conda or pip in the activate environment*
 
 
+
```
ax3l, line 191:
```suggestion
```

services_and_applications/jupyter/overview.rst
```diff
@@ -107,10 +107,13 @@ the Launcher and type ``conda list``. The libraries in the base environment shou
...
 JupyterLab "Launcher" page.
 
-From the Terminal, you can install additional libraries with a simple ``conda install`` to
+From the Jupyter Terminal, you can install additional libraries with a simple ``conda install`` to
 use in your current session. But these installed libraries won't persist across sessions
-if your server is restarted. 
+if your server is restarted.
 
+Since conda's dependency concretizer has sometimes problems to extend existing environments with new
+libraries, you can also use the `mamba <https://github.com/mamba-org/mamba>`_ package manager. Just
```
ax3l, line 115:
```suggestion
libraries, you can also use the `mamba <https://github.com/mamba-org/mamba>`__ package manager. Just
```

services_and_applications/jupyter/overview.rst
```diff
@@ -185,6 +188,22 @@ You may follow these general steps to create and install packages within an any
  - *install packages with conda or pip in the activate environment*
 
 
+Example: Creating a Conda environment for Matplotlib Widgets
+============================================================
+
+We currently install Jupyterlab version 2, which needs additional treatment to get interactive extensions
+`such as matplotlib widgets installed <https://github.com/matplotlib/ipympl>`_.
```
ax3l, line 195:
```suggestion
`such as matplotlib widgets installed <https://github.com/matplotlib/ipympl>`__.
```

services_and_applications/jupyter/overview.rst
```diff
@@ -185,6 +188,22 @@ You may follow these general steps to create and install packages within an any
...
 
+Example: Creating a Conda environment for Matplotlib Widgets
+============================================================
+
+We currently install Jupyterlab version 2, which needs additional treatment to get interactive extensions
+`such as matplotlib widgets installed <https://github.com/matplotlib/ipympl>`__.
+
+Here is an example how to install matplotlib:
+
+::
```
ax3l, line 199:
```suggestion
.. code-block:: bash

```

services_and_applications/jupyter/overview.rst
```diff
@@ -185,6 +188,23 @@ You may follow these general steps to create and install packages within an any
...
+
+Here is an example how to install matplotlib:
+
+.. code-block:: bash
+
+   conda install -c conda-forge ipympl==0.5.8
+   jupyter labextension install @jupyter-widgets/jupyterlab-manager jupyter-matplotlib@0.7.3
+
+Once the installation fishes, restart the kernel of your current notebook and refresh the page.
+You can now use the `%matplotlib ipympl` magic before importing matplotlib to render interactive plots.
```
ax3l, line 205:
```suggestion
You can now use the ``%matplotlib ipympl`` magic before importing matplotlib to render interactive plots.
```

services_and_applications/jupyter/overview.rst
```diff
@@ -185,6 +188,23 @@ You may follow these general steps to create and install packages within an any
...
+
+Here is an example how to install matplotlib:
+
+.. code-block:: bash
+
+   conda install -c conda-forge ipympl==0.5.8
+   jupyter labextension install @jupyter-widgets/jupyterlab-manager jupyter-matplotlib@0.7.3
+
+Once the installation fishes, restart the kernel of your current notebook and refresh the page.
+You can now use the ``%matplotlib ipympl`` magic before importing matplotlib to render interactive plots.
```
ax3l, line 205: Note: active in-support ticket.

After this when running in a non-persistent session, I still get
```
Error displaying widget: model not found
```

For a simple test
```py
%matplotlib ipympl
import matplotlib.pyplot as plt
plt.figure()
plt.show()
```

I think to remember that restarting the kernel is not sufficient, but that jupyter needs to be restarted as well... (which would reset the temporary session)

ax3l: This works now by default.

services_and_applications/jupyter/overview.rst
```diff
@@ -185,6 +188,23 @@ You may follow these general steps to create and install packages within an any
...
+
+
```
ax3l, line 207: Installed now by default
```suggestion
```

ax3l:
@proutrc @kincl  @secondspass ready for review :)

secondspass:
@ax3l can you provide some context on what problems conda has in adding new libraries? It's a little unspecific and not clear what potential problems mamba solves.

ax3l:
@secondspass thanks for the feedback!

Yes, a main challenge is often that the conda default channels ship relatively old software, which are already present in the `base` environment. If people execute `conda install ...` the conda solver itself takes not only very long but also often finds undesired solutions, such as not picking up the latest release of requested software.

This is especially common when installing software from `-c conda-forge` on top of a non-empty, base environment with a couple of existing packages.

secondspass:
That makes sense, can you write something like "since the conda dependency solver can sometimes be slow when adding packages and sometimes not pick up the latest release of a package, you can also use mamba..."? Just basic detail like that should be enough.

PR #348, opened 2021-02-02
Changes for Moderate Enhanced projects

isubsmith:
Adds "citadel" host key.
Adds batch-spi and batch-spi-hm queues to Summit user guide
Adds lots of stuff about file systems. Clarifies when projects get HPSS storage as well (e.g. moderate export control does not have world shared or HPSS)


isubsmith:
Don't merge this until we get the approval from Robert Michael

chrisfuson (approved):

PR #366, opened 2021-04-13
Updating Slate VolumeSnapshot documentation

jlong49:


services_and_applications/slate/examples.rst
```diff
@@ -326,3 +326,20 @@ Basic PVC
...
+.. code-block:: yaml
+
+    apiVersion: snapshot.storage.k8s.io/v1beta1
+    kind: VolumeSnapshot
+    metadata:
+    # Snapshot name
+      name: pvc1-snap
+    spec:
+      source:
+      # Persistent Volume to snapshot
```
kincl, line 343: I would indent these to be even with the line they are referencing

kincl (commented):
We should also include how to "roll back" or provision a new volume from a snapshot

services_and_applications/slate/examples.rst
jlong49: I added the suggestions.

services_and_applications/slate/storage.rst
```diff
@@ -220,6 +207,51 @@ An example Persistent Volume Claim that implements snapshots would look similar
...
+    apiVersion: snapshot.storage.k8s.io/v1beta1
+    kind: VolumeSnapshot
+    metadata:
+      # Name of VolumeSnapshot
+      name: pvc-snap
+    spec:
+      source:
+        # Name of persistentVolumeClaim to snapshot
+        persistentVolumeClaimName: snapshot-pvc
+      volumeSnapshotClassName: csi-snapclass
```
kincl, line 224: This VolumeSnapshotClass name seems kind of arbitrary from a user perspective. If we don't have any "options" we need to provide trident with this maybe we can name it `netapp-default` and then make it the default storage class with the annotation `snapshot.storage.kubernetes.io/is-default-class: "true"`

jlong49: I added a new VolumeSnapshotClass to the clusters named netapp-default and removed the default annotation on the existing csi-snapclass VolumeSnapshotClass. I left the csi-snapclass in place since there is at least one user on Marble that is using VolumeSnapshots with the csi-snapclass VolumeSnapshotClass.

kincl (approved):

PR #375, opened 2021-04-30
update spock quick start after change to slurm

tom-papatheodore:
The spock quick-start needed to be updated after changes to slurm's default for `--thread-per-core`


isubsmith (approved):
Text looks good to me. I didn't verify all the examples though.

PR #376, opened 2021-05-15
Updated GPU mapping section of Spock quick-start

tom-papatheodore:
I updated the GPU Mapping section of the Spock user guide to use `--gpus-per-task` for 1-MPI-per-GPU examples and `--ntasks-per-gpu` for multiple-MPI-per-GPU examples. I also added 2-node versions of each example as well.

### DIFF
--- a/systems/spock_quick_start_guide.rst
+++ b/systems/spock_quick_start_guide.rst
@@ -279,7 +279,7 @@ To use GPU-aware Cray MPICH with the Cray compiler wrappers, users must load spe
 
 .. note:: 
 
-    Setting ``MPICH_SMP_SINGLE_COPY_MODE=CMA`` is required as a temporary workaround due to a known issue. Users should make a note of where they set this environment variable (if e.g., set in a script) since it should NOT be set once the known issue has been resolved.
+    Setting ``MPICH_SMP_SINGLE_COPY_MODE=CMA`` is required as a temporary workaround due to a `known issue <https://docs.olcf.ornl.gov/systems/spock_quick_start_guide.html#olcfdev-138-gpu-aware-cray-mpich-can-cause-hang-in-some-codes>`__. Users should make a note of where they set this environment variable (if e.g., set in a script) since it should NOT be set once the known issue has been resolved.
 
 .. code:: bash
 
--- a/systems/spock_quick_start_guide.rst
+++ b/systems/spock_quick_start_guide.rst
@@ -37,6 +37,14 @@ via Infinity Fabric (xGMI), allowing for a peak device-to-device bandwidth of
    :width: 100%
    :alt: Spock node architecture diagram
 
+.. note::
+    There are 4 NUMA domains per node, that are defined as follows:
+
+    * NUMA 0: hardware threads 000-015, 064-079 | GPU 0
+    * NUMA 1: hardware threads 016-031, 080-095 | GPU 1
+    * NUMA 2: hardware threads 032-047, 096-111 | GPU 2
+    * NUMA 3: hardware threads 048-063, 112-127 | GPU 3
+
 System Interconnect
 -------------------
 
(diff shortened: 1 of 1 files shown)


tom-papatheodore:
I added an unrelated commit that simply adds an extra MPICH env var needed for GPU-Aware MPI.

reubendb (commented):
I think "workaround" used as a noun should be one word (e.g. "a workaround" instead of "a work around" as currently in Example 6 (e.g. https://en.wikipedia.org/wiki/Workaround).
Otherwise, looks good I think.

systems/spock_quick_start_guide.rst
```diff
@@ -282,6 +290,7 @@ To use GPU-aware Cray MPICH with the Cray compiler wrappers, users must load spe
     ## These must be set before running
     export MPIR_CVAR_GPU_EAGER_DEVICE_MEM=0
     export MPICH_GPU_SUPPORT_ENABLED=1
+    export MPICH_SMP_SINGLE_COPY_MODE=CMA
```
verolero86, line 293: For this variable, we should mention that it is needed temporarily due to a known issue as we do want people to remove this once the issue is fixed

systems/spock_quick_start_guide.rst
```diff
@@ -777,59 +828,155 @@ The ordering of the GPU mapping can also be changed with ``--gpu-bind=map_gpu``
...
+
+The output shows the round-robin (``cyclic``) distribution of MPI ranks to GPUs. In fact, it is a round-robin distribution of MPI ranks *to NUMA domains* (the default distribution). The GPU mapping is a consequence of where the MPI ranks are distributed; ``--gpu-bind=closest`` simply maps the GPU in a NUMA domain to the MPI ranks in the same NUMA domain.
+
+**Example 6: 16 MPI ranks - where 2 ranks share a GPU (round-robin, multi-node)**
 
+This example is an extension of Example 5 to run on 2 nodes.
+
+.. warning::
+
+    This example requires a work around to run as expected. ``--ntasks-per-gpu=2`` does not force MPI ranks 008-015 to run on the second node, so the number of physical CPU cores per MPI rank is increased to 8 (``-c8``) to force the desired behavior due to the constraint of the number of physical CPU cores (64) on a node.
```
verolero86, line 899: adding @reubendb 's comment: work around -> workaround

systems/spock_quick_start_guide.rst
```diff
@@ -777,59 +828,155 @@ The ordering of the GPU mapping can also be changed with ``--gpu-bind=map_gpu``
...
 
-    On AMD's MI100 GPUs, multi-process service (MPS) is not needed since multiple MPI ranks per GPU is supported natively.
+    In the job step for this example, ``--distribution=*:block`` is used, where ``*`` represents the default value of ``block`` for the distribution of MPI ranks across compute nodes and the distribution of MPI ranks across NUMA domains has been changed to ``block`` from its default value of ``cyclic``.
+
+.. note:: 
+
+    Because the distribution across NUMA domains has been changed to a "packed" (``block``) configuration, caution must be taken to ensure MPI ranks end up in the NUMA domains where the GPUs they intend to be mapped to are located. To accomplish this, the number of physical CPU cores assigned to an MPI rank was increased - in this case to 8. Doing so ensures that only 2 MPI ranks can fit into a single NUMA domain. If the value of ``-c`` was left at ``1``, all 8 MPI ranks would be "packed" into the first NUMA domain, where the "closest" GPU would be GPU 0 - the only GPU in that NUMA domain. 
+
+    Notice that this is not a work around like in Example 6, but a requirement due to the ``block`` distribution of MPI ranks across NUMA domains.
```
verolero86, line 937: work around -> workaround

verolero86 (changes requested):
Thank you @tom-papatheodore ! These look great!

tom-papatheodore:
Ok, the requested changes have been made.

systems/spock_quick_start_guide.rst
reubendb: Should a new "Known Issue" be added (and cross-referenced, perhaps) for this issue requiring the workaround?

tom-papatheodore: @reubendb Yes. I've now added a "known issue with workaround" but had to guess at the link name - since it is dynamically created from the JIRA known issues. It looks like a straightforward format it follows, but we won't be able to test it until it is live (since that part is only available in the production docs).

verolero86 (approved):

PR #384, opened 2021-06-07
NVHPC compiler additions to the compiler section

secondspass:
Added options for the Nvidia HPC SDK's set of compilers in the Summit guide's compiler section. They're the PGI compilers in new packaging.


isubsmith (approved):
LGTM

PR #404, opened 2021-08-06
Moving data section to single page

warenaud:


secondspass (approved):
I built it and checked. Looks all good to me.

secondspass:
Hm, hang on. There seems to be something up with the rendering of some of the tables. Checking.

secondspass (approved):
Never mind, all is good.

PR #409, opened 2021-08-19
Summit myolcf update

warenaud:
Was adding a myOLCF link to the Summit page when I noticed the discussions of job accounting differed slightly between the Summit and Andes guides, and the Andes guide had an old link for myOLCF. Modified to make the two pages consistent, to link to the new myOLCF, and to update references form 'My OLCF' to the newer 'myOLCF' styling.


isubsmith (approved):
LGTM

PR #410, opened 2021-08-21
spock: update GPU-aware MPI instructions

jedbrown:
Directory /opt/cray/pe/mpich/8.1.4 no longer exists.


ddietz89:
@jedbrown Good catch! Thank you for sending this up. Those variables shouldn't be necessary at all now, so I'm going to push up another commit that removes them entirely.

PR #432, opened 2021-10-27
restructuring multithreading information for Summit jobs

secondspass:
I've added a section explaining the use of OMP_NUM_THREADS. I've reworked the docs a little so that the Job Launcher (jsrun) section is broken out into more pieces to give more visibility to the subsections that were usually hidden under Job Launcher (jsrun) and were hard to find.

You will want to pull my branch locally and build it with sphinx to see how it looks on the rendered webpage.

### DIFF
--- a/systems/summit_user_guide.rst
+++ b/systems/summit_user_guide.rst
@@ -1463,17 +1463,17 @@ As on any system, it is useful to keep in mind the hardware underneath every
 execution. This is particularly true when laying out resource sets.
 
 Launching a Job with jsrun
-^^^^^^^^^^^^^^^^^^^^^^^^^^
+--------------------------
 
 jsrun Format
-""""""""""""
+^^^^^^^^^^^^
 
 ::
 
       jsrun    [ -n #resource sets ]   [tasks, threads, and GPUs within each resource set]   program [ program args ]
 
 Common jsrun Options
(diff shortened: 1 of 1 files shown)


systems/summit_user_guide.rst
```diff
@@ -1662,8 +1662,196 @@ contain 6 MPI tasks (``-a6``), 3 GPUs (``-g3``), and 6 cores
...
+explicit and set the ``OMP_NUM_THREADS`` value yourself rather than
+relying on Jsrun constructing it for you. Especially when you are
+using `Job Step Viewer`_ which relies on the presence of that
+environment variable to give you visual thread assignment information.
+
+In the below example, you could also do ``export OMP_NUM_THREADS=16`` in your
+job script instead of passing it as a ``-E`` flag to jsrun. The below example
+starts 1 resource set with 2 tasks and 8 cores, 4 cores bound to each task,
+16 threads for each task.
+
```
michael-sandoval, line 1805: Since you say below this why you can't do 32 threads, I would include somewhere around here why you're able to do 16 threads -- a.k.a 4 * 4  = 16 because SMT4 and 4 cores

michael-sandoval (changes requested):
Looks good to me.

michael-sandoval (approved):

PR #438, opened 2021-11-16
Update cupy.rst

gounley:
A natural modification of the cupy install instructions is to use them with the open-ce module. However, while open-ce loads the cuda module automatically, it does not set the CUDA_PATH variable when loaded this way. Adding a line to clarify this point.


michael-sandoval (approved):
Thanks for catching this!

PR #452, opened 2022-01-13
Update crusher_quick_start_guide.rst

nicholasmalaya:
AMD recommends changing from 52 TF to 53 TF. This is the correct 'peak performance of 53 TFLOPS in double-precision for modeling and simulation'.


verolero86 (approved):
Thank you @nicholasmalaya !

PR #453, opened 2022-01-15
corrected MI250X count to 768 instead of 1536 (which is GCD count)

tom-papatheodore:
The number of MI250X in Crusher was accidentally listed as 1536 (which is the number of GCDs not MI250X). This PR changes that number to be the correct value of 768.

192 compute nodes * (4 MI250X / compute node) = 768 MI250X
192 compute nodes * (4 MI250X / compute node) * (2 GCDs / MI250X) = 1536 GCDs


tom-papatheodore:
I merged this PR myself since it was a simple change from 1536 MI250X to 768 MI250X.

PR #454, opened 2022-01-18
Updated the outdated OpenCE page

michael-sandoval:
Removed references and examples of deprecated modules and commands


LeahHuk (approved):
Looks good to me

PR #459, opened 2022-01-25
MPICH and ROCm compatibility on Crusher

mttjdvs:
Compatibility between cray-mpich and rocm is not guaranteed across multiple versions. This commit adds documentation to help the user determine a compatible pairing of cray-mpich and rocm.

### DIFF
--- a/systems/crusher_quick_start_guide.rst
+++ b/systems/crusher_quick_start_guide.rst
@@ -249,6 +249,11 @@ To use GPU-aware Cray MPICH, users must set the following modules and environmen
 
     export MPICH_GPU_SUPPORT_ENABLED=1
 
+Determining the Compatibility of Cray MPICH and ROCm
+""""""""""""""""""""""""""""""""""""""""""""""""""""
+
+Releases of ``cray-mpich`` are each built with a specific version of ROCm, and compatibility across multiple versions is not guaranteed. OLCF will maintain compatible default modules when possible. If using non-default modules, you can determine compatibility by reviewing the *Product and OS Dependencies* section in the ``cray-mpich`` release notes. This can displayed by running ``module show cray-mpich/<version>``. If the notes indicate compatibility with *AMD ROCM X.Y or later*, only use ``rocm/X.Y.Z`` modules. If using a non-default version of ``cray-mpich``, you must add ``${CRAY_MPICH_ROOTDIR}/gtl/lib`` to either your ``LD_LIBRARY_PATH`` at run time or your executable's rpath at build time.
+
 .. note::
 
     There are extra steps needed to enable GPU-aware MPI on Crusher, which depend on the compiler that is used (see 1. and 2. below).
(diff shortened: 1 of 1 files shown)


mttjdvs:
This may be one way to accomplish this. It's intentionally lacking on some specific information that may change with updates to the programming environment and rocm modules. I had initially intended to add a table marking compatible pairings of cray-mpich and rocm, but decided against it since that would need require extra maintenance with new releases and the potential for incorrect docs. If the clarity that a table provides outweighs those concerns then I can certainly add it here.

Does the relative position of this addition make sense? I had also considered adding it near the end of the GPU-aware section as well.

tom-papatheodore:
IMO, this does belong at the end of the GPU-aware section. Otherwise, a user might get to this part and then decide to just trying themselves without reading the (more crucial) sub-sections 1 and 2. 

Regarding the content itself, I have
* There is a missing "be" in the sentence "This can displayed by running"
* Should there be an "or later" at the very end of this sentence? "If the notes indicate compatibility with *AMD ROCM X.Y or later*, only use ``rocm/X.Y.Z`` modules."
* For the last sentence, why is it necessary to add the GTL libs to `LD_LIBRARY_PATH`? `PE_MPICH_GTL_DIR_amd_gfx90a` appears to get updated when I change from `cray-mpich/8.1.12` to `cray-mpich/8.1.11`

mttjdvs:
> IMO, this does belong at the end of the GPU-aware section. Otherwise, a user might get to this part and then decide to just trying themselves without reading the (more crucial) sub-sections 1 and 2.
> 
> Regarding the content itself, I have
> 
> * There is a missing "be" in the sentence "This can displayed by running"

Agreed. I'll get these issues fixed.

> * Should there be an "or later" at the very end of this sentence? "If the notes indicate compatibility with _AMD ROCM X.Y or later_, only use `rocm/X.Y.Z` modules."

The "or later" is in fact intentionally left out. Specifying "or later" is a mistake on Cray's part, since AMD does not guarantee this compatibility across releases. Cray seems to have fixed this wording in the latest release notes. If they maintain the proper wording in future releases then I think this sentence can be removed when/if the older PEs are removed from the system.

> * For the last sentence, why is it necessary to add the GTL libs to `LD_LIBRARY_PATH`? `PE_MPICH_GTL_DIR_amd_gfx90a` appears to get updated when I change from `cray-mpich/8.1.12` to `cray-mpich/8.1.11`

PE_MPICH_GTL_DIR_amd_gfx90a and PE_MPICH_GTL_LIBS_amd_gfx90a are only used for linking during the build. When it comes time to run, the GTL library is found with ld.so.cache unless the user provides a way for ld.so to find it sooner. That is done by setting LD_LIBRARY_PATH at runtime or the executable's rpath/runpath during the build.

tom-papatheodore:
Gotcha! Thanks. Then, pending @verolero86's review, this is approved by me.

tom-papatheodore (approved):
See my previous comments.

verolero86 (approved):
Looks good to me!

PR #464, opened 2022-02-04
Adding 'Differences between Summit and Crusher' Section

nicholasmalaya:
Update to Crusher Quick Start guide. 

  * Adding a section that documents important differences between Summit and Crusher. 
  * To be populated with more information in the future.
  * Initially documenting differences in denorm, 'flush to zero' behavior on MI250X and V100.


verolero86 (approved):
Looks great - thank you @nicholasmalaya !

PR #465, opened 2022-02-05
Initial XNACK and page migration section

abbotts:
This PR adds information on page migration on Crusher and the impact of XNACK settings.

There's certainly some missing and possibly incorrect information here, but this should serve as a start we can refine.


tom-papatheodore:
The flow of this makes sense to me; start with the XNACK mode (`HSA_XNACK`) since default is `xnack any`. Then put the compiling options at the bottom for possible performance improvements. 

Do you think the following items should be clarified? 

- Compiling with `xnack+` while `HSA_XNACK` is set to 0 will result in "hipErrorNoBinaryForGpu: Unable to find code object for all current devices!" 

- The entries in the table describe the behavior of the current processor when "accessing memory last touched by the _other_ processor". 

Otherwise, I think the content looks great.

abbotts:
I think clarifying the error modes is a good suggestion. I'll add that.

>The entries in the table describe the behavior of the current processor when "accessing memory last touched by the other processor".

You're right that we should clarify the tables more. Really they describe behavior after the *first GPU access*, since the implementation is biased towards GPU and will not migrate pages back to DDR4 unless forced to.

I'll make these changes, then probably convert this to a proper PR and add reviewers. Thanks @tom-papatheodore

abbotts:
>The entries in the table describe the behavior of the current processor when "accessing memory last touched by the other processor".

I'm still not happy with how I worded things in the table. I'm open to suggestions for sure.

tom-papatheodore:
This LGTM. I suggest we publish this to get the info into the users' hands. We can add code examples later. 
@abbotts, @verolero86 Any objections to me merging this?

verolero86 (approved):
Looks good to me!

PR #482, opened 2022-02-24
VisIt: troubleshoot and module updates

michael-sandoval:
Added syntax for module usage on Andes

Added troubleshooting for hanging connections


LeahHuk (approved):
lgtm

PR #485, opened 2022-03-02
adding containers on summit documentation

secondspass:


proutrc (commented):
Is line 79 (https://github.com/olcf/olcf-user-docs/pull/485/files#diff-1497ac665fded6e44ca352186e9a268b19e65d6c41ff5a1e86a63561eac0858aR79) suppose to be there or have more?

One other thing - do we need to outline the cgroup modification request?

secondspass:
@proutrc fixed the hanging sentence.

re: cgroup modification request. Right now every user's cgroup memory limit has been increased to 32gb. Which _should_ be enough for most cases. Don wants to make sure he does as little work as possible, hence the blanket increase. We could tolerate a user running singularity build on the compute node since that doesn't need the internet. But if they come to the point where they have to run 'podman build' on the compute node because they need the memory, then we need to reevaluate the cgroup limits.

proutrc (approved):
LGTM

PR #497, opened 2022-03-22
rework gpu binding examples on crusher

mttjdvs:
The --gpus-per-task option is currently difficult to use due to how Slurm works on Crusher's compute nodes. This commit reworks the current examples to use either --gpus or --gpus-per-node to get the desired binding. An example of how to run the test case of a single task with a single GPU is also included.

### DIFF
--- a/systems/crusher_quick_start_guide.rst
+++ b/systems/crusher_quick_start_guide.rst
@@ -655,12 +655,12 @@ Now the output shows that each OpenMP thread ran on (one of the hardware threads
 GPU Mapping
 ^^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


tom-papatheodore:
I made a couple of minor changes but otherwise this LGTM. Thanks for fixing this section!

While we're updating this section, should we also edit the "Multiple GPUs per MPI rank" section? It currently reads:

> **Multiple GPUs per MPI rank**
> As mentioned previously, all GPUs are accessible by all MPI ranks by default, so it is possible to programatically map any combination of MPI ranks to GPUs.

So at a minimum we should correct this sentence to read "any combination of GPUs to MPI ranks" - since that's what the section is about : )

But should we also suggest *not* mapping more than 1 GPU to a single MPI rank? And possibly even note that this is currently not supported with Cray's GPU-aware MPICH?

mttjdvs:
Thanks for the fixes and suggestion. I've added the language for multiple GPUs per rank.

tom-papatheodore (approved):
@mttjdvs Ok, I'm going to approve this so you can submit when ready. Thanks again for re-working this section.

tom-papatheodore:
...err, merge when ready : )

PR #498, opened 2022-03-23
Update --threads-per-core docs for crusher 

mttjdvs:
When using --threads-per-core=<threads> with srun, the --threads-per-core value set with salloc/sbatch must be at least as large as the value in srun.

This was good practice before, but has now been made an explicit requirement as of Slurm 21.08.6


tom-papatheodore (approved):
Hey @mttjdvs. I made a couple of minor edits to this PR, but otherwise it LGTM.

PR #504, opened 2022-04-12
Add compatibility table for rocm and cray mpich

mttjdvs:
This adds a table for users to track the compatibility of rocm and cray-mpich modules on crusher. This makes the above text on determining compatibility more concrete.


verolero86 (approved):
LGTM! 👍

tom-papatheodore (approved):
This LGTM. Technically, `cray-mpich/8.1.15` only shows ROCM 5.0 support but as you mentioned, you tested the compatibility with linker testing - so I think putting ROCM 5.1.0 in the table makes sense. IMO, merge away!

PR #507, opened 2022-04-20
Update HIP module info on Summit

mttjdvs:
The name and behavior of the HIP module on Summit has changed.

There is probably more that can be said about what this release offers, specifically in regards to the new HIP marshalling libraries used for the cuda math libraries. That might be a larger documentation effort though, and I'm not sure it should delay correcting this piece now.


tom-papatheodore (approved):
The content on this looks good to me, although I didn't actually try to build this locally.

verolero86 (approved):
Thanks for fixing it to match the new modulefile @mttjdvs !

PR #508, opened 2022-04-20
Added section for profiling and roofline profiling

hagertnl:
Added a section to detail some very basic rocprof usage - 'rocprof --stats'. Also added a section about how to generate a simple operational roofline model using rocprof metrics. Fairly limited example (ignores matrix instructions), but it should be able to generate correct rooflines. Rocprof's timestamps aren't fully accurate at this moment (BeginNs is not appearing to be correct), so maybe adding a warning to the doc is needed as well.


verolero86 (approved):
LGTM! thank you @hagertnl 👍

PR #515, opened 2022-05-18
Update crusher mpich and rocm compat table

mttjdvs:
Updating the compatibility table for the recent crusher upgrade.


tom-papatheodore (approved):
The content looks correct to me. I also built and viewed the docs locally and it looks good.

PR #522, opened 2022-05-25
Update overview.rst

ketancmaheshwari:
Adding link to workflows docs


michael-sandoval (changes requested):
Good idea to link to the other section, but probably need to throw in more than just the link. Something like "Scientific workflows at OLCF can be found in the :ref:`workflows` page under the Software section."

PR #525, opened 2022-06-07
Adding sections for GPU containers

secondspass:
Added two sections in Containers on Summit talking about single node
GPU program and a GPU aware MPI program. Both work on Summit if the steps are
followed.

Also added info about the seg fault you'll see if you don't use the right
flag for GPU aware MPI, and corrected some inconsisten capitalization in the
Summit guide.


michael-sandoval (approved):
LGTM

PR #533, opened 2022-06-29
Fixed typo in RDREQ and WRREQ metrics of rocprof section

hagertnl:
Fixed a typo in the body of paragraphs that referred to L2 write/read request as WREQ and RREQ, instead of their correct naming - WRREQ, RDREQ.


tom-papatheodore (approved):
This LGTM. I saw the ticket and this appears to fix the typos.

PR #541, opened 2022-08-03
Added a section on using CrayPat and HPCToolkit.

jholmen:
Added a section on using the HPE Performance Analysis Tools.


hagertnl:
It might be good to include mention of the former name for this package -- craypat, and to include the acronym PAT in the section heading (ie, "HPE Performance Analysis Tools (PAT)"). It also could be good to have outlined examples for using PrgEnv-cray+CC, PrgEnv-amd+CC, PrgEnv-amd+hipcc. I'd think these are likely the most common compiler setups, so it might be good to show them

chrisfuson (approved):

PR #547, opened 2022-08-11
additional details to AMD section. Need AMD to verify

secondspass:


tom-papatheodore (approved):
LGTM. I'll point AMD to this to check for correctness.

PR #548, opened 2022-08-11
Update Roofline Profiling Equations in frontier_user_guide.rst

nicholasmalaya:
Feel free to reject, but I'd like to propose adding \text to make equations in roofline profiling section more clear. This also makes the subscripts render correctly.


michael-sandoval (approved):
Good idea, thanks!

PR #549, opened 2022-08-11
Replace "Summit" with "Frontier" in text

nicholasmalaya:
Replacing "Summit" (OLCF-4) with "Frontier" (OLCF-5). Two instances were found and changed.


frobnitzem (approved):
Thanks!

michael-sandoval (approved):
Thanks, we also plan to change the wording (and remove launch nodes). That section was accidentally copy pasted from Summit's guide

PR #551, opened 2022-08-15
Corrected the table of node types in overview section

LeahHuk:
Removed the reference to launch nodes. Updated the jsrun reference to srun.


michael-sandoval (approved):
LGTM

PR #556, opened 2022-08-17
Touch up the AMD GPUs section

jakurzak:


systems/frontier_user_guide.rst
```diff
@@ -233,54 +233,48 @@ And here is the output from the script:
...
+    The Slurm workload manager and the ROCr runtime treat each GCD as a separate GPU
+    and visibility can be controlled using the ``ROCR_VISIBLE_DEVICES`` environment variable.
+    Therefore, from this point on, the Frontier guide simply refers to a GCD as a GPU.
 
-Each GPU has a peak performance of 26.5 TFLOPS (double-precision), 110 compute units, and
-64 GB of high-bandwidth memory (HBM2) which can be accessed at a peak of 1.6 TB/s. The 2
-GPUs on an MI250X are connected with [4x] GPU-to-GPU Infinity Fabric links for a total
-bandwidth of 200+200 GB/s . Consult the diagram in the  section :ref:`frontier-nodes` for
-information on how the accelerators connect to each other, to the CPU, and to the network.
+Each GPU contains 110 Compute Units (CUs) grouped in 4 Compute Engines (CEs).
```
secondspass, line 250: HI @jakurzak , one thing I wanted to clarify is the fact that each GCD has 110 CUs but the diagram you added shows 112 CUs in each GCD. Can you clarify if there are two that are inactive and which ones they are?

systems/frontier_user_guide.rst
```diff
@@ -233,54 +233,48 @@ And here is the output from the script:
...
+A command processor in each GPU receives API commands and transforms them into compute tasks.
+Compute tasks are managed by the 4 compute engines, which dispatch wavefronts to compute units.
+All wavefronts from a single workgroup are assigned to the same CU.
+In CUDA terminology, workgroups are "blocks", wavefronts are "warps", and work-items are "threads".
```
secondspass, line 254: There is already a section just below this section addressing AMD vs Nvidia terminology with a table. So saying it here may be redundant.

secondspass (changes requested):

systems/frontier_user_guide.rst
jakurzak: I clarified that there are 112 physical CUs and two are disabled.

systems/frontier_user_guide.rst
jakurzak: I feel like it does not hurt to reiterate, but let me know otherwise.

systems/frontier_user_guide.rst
```diff
@@ -248,6 +248,7 @@ A single Frontier node contains 4 MI250X OAMs for the total of 8 GCDs.
     Therefore, from this point on, the Frontier guide simply refers to a GCD as a GPU.
 
 Each GPU contains 110 Compute Units (CUs) grouped in 4 Compute Engines (CEs).
+Physically, each GPU contains 112 CUs, but two are disabled for yield reasons.
```
secondspass, line 251: Hey Jakub, can you clarify what you mean when you say yield reasons? I'm not sure that's clear.

secondspass (changes requested):

systems/frontier_user_guide.rst
jakurzak: Okay, I removed "yield reasons".
There's no good reason confusing the reader with semiconductor manufacturing issues.
Seems beyond the scope of the guide.

secondspass (approved):

PR #559, opened 2022-08-23
Initial add of SPI section to Summit docs

chrisfuson:


systems/summit_user_guide.rst
```diff
@@ -4293,6 +4293,97 @@ judiciously, especially when running the debugger in "batch" or
...
+========================
+
+.. _enhanced-enclave:
+
+Enhanced Enclave
+------------
+The OLCF provides multiple security enclaves to support multiple levels of computing. 
+
+* Open
+  * Training resources such as Andes is an example
```
LeahHuk, line 4309: Did you mean Ascent?

LeahHuk (changes requested):
I assume its meant to be bulleted skeleton for now as a jumping off point. I only saw the one issue with the Open enclave section where I think you meant to say Ascent and not Andes. I can work on turning the bullet points into actual text.

systems/summit_user_guide.rst
chrisfuson: Yes; fixed

chrisfuson:
> 

I can format differently as the next step if bulleted is not the best in this case.

LeahHuk (approved):

PR #567, opened 2022-09-14
Added to rocprof roofline profiling section

hagertnl:
Added:

- TRANS op to FP64_FLOPS calculation
- Usage example of rocprof with >1 MPI ranks
- Calculation of theoretical flops on crusher
- Added note for --timestamp on bug in the rocprof section


tom-papatheodore (approved):
LGTM!

PR #577, opened 2022-10-11
Add HPCToolkit section on Summit user guide

wyphan:
The added section points to the modulefile for the newest HPCToolkit release version 2022.10.01.
Tested locally on my laptop.

### DIFF
--- a/systems/summit_user_guide.rst
+++ b/systems/summit_user_guide.rst
@@ -2805,6 +2805,37 @@ trace files for Vampir to visualize.
 For detailed information about using Vampir on Summit and the builds available,
 please see the `Vampir Software Page <https://www.olcf.ornl.gov/software_package/vampir/>`__.
 
+HPCToolkit
+----------
+
+`HPCToolkit <http://hpctoolkit.org/>`__ is an integrated suite of tools for measurement and analysis of program performance on computers ranging from multicore desktop systems to the nation’s largest supercomputers. HPCToolkit provides accurate measurements of a program’s work, resource consumption, and inefficiency, correlates these metrics with the program’s source code, works with multilingual, fully optimized binaries, has very low measurement overhead, and scales to large parallel systems. HPCToolkit’s measurements provide support for analyzing a program execution cost, inefficiency, and scaling characteristics both within and across nodes of a parallel system.
+
+Programming models supported by HPCToolkit include MPI, OpenMP, OpenACC, CUDA, OpenCL, DPC++, HIP, RAJA, Kokkos, and others.
+
+Below is an example that generates a profile and loads the results in their GUI-based viewer.
+
+.. code:: bash
+
+    module use /gpfs/alpine/csc322/world-shared/modulefiles/ppc64le
+    module load hpctoolkit/2022.10.01
+
+    # 1. Profile and trace an application using CPU time and GPU performance counters
+    jsrun <jsrun_options> hpcrun -o <measurement_dir> -t -e CPUTIME -e gpu=nvidia <application>
+
+    # 2. Analyze the binary of executables and its dependent libraries
+    hpcstruct <measurement_dir>
+
+    # 3. Combine measurements with program structure information and generate a database
+    hpcprof -o <database_dir> <measurement_dir>
+
+    # 4. Understand performance issues by analyzing profiles and traces with the GUI
+    hpcviewer <database_dir>
+
+A quick summary of HPCToolkit options can be found in the `HPCTookit wiki page <https://gitlab.com/hpctoolkit/hpctoolkit/-/wikis/home>`__. More detailed information on HPCToolkit can be found in the `HPCToolkit User's Manual <http://hpctoolkit.org/manual/HPCToolkit-users-manual.pdf>`__.
+
+.. note::
+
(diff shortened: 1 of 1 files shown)


michael-sandoval (changes requested):
Heard from the software team that a proper module is going to be added sometime this week, so will need to update the `module use` and maybe the `module load` line when that happens. Otherwise looks good

michael-sandoval:
This will probably be mentioned in a relevant ticket of yours, but putting it here as well.

hpctoolkit/2022.10.01 and hpctoolkit/2022.10.01-cpu are now available on Summit for GCC toolchain environments:

```
$ module spider hpctoolkit/2022.10.01 

------------------------------------
  hpctoolkit: hpctoolkit/2022.10.01
------------------------------------

      gcc/10.2.0  spectrum-mpi/10.4.0.3-20210112
      gcc/11.1.0  spectrum-mpi/10.4.0.3-20210112
      gcc/11.2.0  spectrum-mpi/10.4.0.3-20210112
      gcc/9.1.0  spectrum-mpi/10.4.0.3-20210112
      gcc/9.3.0  spectrum-mpi/10.4.0.3-20210112
      spectrum-mpi/10.4.0.3-20210112
```

So you should be good to make the changes now

wyphan:
@michael-sandoval The problem with this deployment (GCC toolchain environments only) is that although HPCToolkit can only be compiled using GCC at the moment, the resulting installation is usable for users compiling their code with _any_ compiler toolchain. The way Lmod modules are set up on Summit for compilers prevents having more than one compiler toolchain at the same time because the modules belong in the same module family. From our current application engagement activities, HPCToolkit users on Summit tend to prefer NVHPC because it produces the most performant code for the NVIDIA V100 GPUs on Summit.

michael-sandoval:
@wyphan Ahhhh, I see. Would you have any objections to the HPCToolkit builds we expose under all compiler environments being built with the OS gcc v8 and cuda/11.0.3? That's the only way we can really make that option to play nicely with the module system is if we end up building with the OS's versions

michael-sandoval:
Essentially, we're trying to avoid a situation of offering a "core" module (a module that avoids the Lmod toolchain dependencies) on our system/docs which isn't guaranteed to work with every toolchain option/cuda version that a user might use. Using the OS's built in GCC and the default cuda module to build HPCToolkit would be as close to "guaranteeing" that as possible. Basically, I'm asking you - as a dev - if you can confirm that no matter what HPCToolkit is built with, it'll still work for users running codes with different library versions that their code is built with that they may be linking instead.

wyphan:
@michael-sandoval GCC 8 is not a problem, but the CUDA version is a problem because CUDA >= 11.5 organizes things differently from 11.0.3. Perhaps @jmellorcrummey can chime in a lil bit.

jmellorcrummey:
@michael-sandoval If built with the system gcc, HPCToolkit will work with code compiled with any compiler (e.g., IBM xlc, NVIDIA nvcc, gcc, ...). As long as HPCToolkit is built without PAPI (`~papi`), if built with any CUDA version 11+, it will work with any CUDA version 11+, even if the version a user uses for their application differs from the version that HPCToolkit was built with. (PAPI binds to a CUDA version at compile time. HPCToolkit uses the CUDA version used by the application it is measuring. Thus, using PAPI breaks cross-version compatibility for HPCToolkit.)

Wil mentioned that we recommend building and measuring applications with CUDA 11.5+. CUDA 11.5+ only offers a benefit for HPCToolkit when someone is using NVIDIA's PC sampling capability to collect measurement data within kernels. The reason 11.5+ helps is that NVIDIA's compilers are improved in 11.5 and later so that they record inlining information about code in GPU kernels. Having inlining info within GPU kernels helps HPCToolkit better attribute performance to complex user codes. If run with a version of CUDA < 11.5, HPCToolkit will still faithfully associate metrics with source lines; however, you may not be able to understand why those source lines have measurement data associated with them if no inlining info is present (because nvcc 11.4 and earlier don't record inlining info in GPU kernels.)

michael-sandoval:
Thanks for the info, seems promising! I'll pass this along.

wyphan:
FYI, we (the HPCToolkit team) maintain our Spack configs for HPCToolkit deployments in [a GitHub repo](https://github.com/jmellorcrummey/spack-configs/), in case the software team at OLCF would like to see what we're doing differently with our builds.

michael-sandoval:
Yup, we are aware (well, the software team is). The last description they seemed to glob onto, so I'll let you know when we have a build in place

wyphan:
Oh, one more thing: the `+mpi` variant in HPCToolkit's Spack recipe (which builds `hpcprof-mpi`) is **disabled** for the 2022.10.01 release version, because we're still working on it (notably, quality assurance and checking for corner cases). However, the regular `hpcprof` is now multithreaded.

We've also noticed a runtime problem with `hpcrun` when built with both `+cuda` and `+papi` on Perlmutter (`cuptiSubscribe` called twice; the first time from `hpcrun` initialization and the second time during initialization of PAPI CUDA component). While we investigate the problem, please disable `+papi` for GPU builds.

michael-sandoval (approved):

PR #586, opened 2022-11-04
updates to the Deployment tutorial

secondspass:
Seems like some more syntax changed in openshift. I've updated that but need someone to check.


kellenrl (approved):
looks good 👍

PR #587, opened 2022-11-28
HW LDS FP32 atomicAdd guidance for Crusher

afanfa:
Guidance on potential low performance of HW LDS FP32 atomicAdd() on MI200.


afanfa:
We haven't fully characterized the impact of this HW behavior. It is good to recommend the users to experiment based on their app access pattern.

tom-papatheodore (approved):
With the suggested fixes, this looks good to me.

PR #591, opened 2022-12-13
pointed Crusher quick-start to correct node diagram

tom-papatheodore:
This PR simply points to the updated Frontier node diagram that includes the 4 IF links between the GCDs on a common MI250X instead of a single darker red line. Until this PR is merged, you can see the difference between the Crusher and Frontier node diagrams in their user guides. 

The `diff` is literally:
```
-.. image:: /images/Crusher_Node_Diagram.jpg
+.. image:: /images/Frontier_Node_Diagram.jpg
```


hagertnl (approved):
lgtm!

PR #601, opened 2022-12-20
Adding Citadel system

chrisfuson:
add Citadel to system user guide list


michael-sandoval (changes requested):
Probably best to have an explicit clarification at the end (line 25) indicating that you're talking about the SPI section/page (even though the hyperlink is correct).

michael-sandoval (approved):
LGTM

PR #602, opened 2022-12-27
Add upcoming system changes on Crusher

verolero86:


systems/crusher_quick_start_guide.rst
```diff
@@ -1702,6 +1702,21 @@ The aforementioned behavior is only true for FP atomicAdd() operations. Hardware
...
+
+System Updates 
+============== 
+
+2022-12-29 
+---------- 
+
+On Thursday, December 29, 2022 the following system configuration settings will be updated on Crusher:
+
+ * Core specialization will be enabled: as a result, system processes will be constrained to core 0 on every node.
```
hagertnl, line 1715: > system processes will be constrained to core 0 on every node

With `-S 8`, wouldn't the processes be constrained to the first core in each NUMA (depending on distribution)? Specifying core 0 feels like it's just core 0 every time, potentially.

hagertnl (approved):
LGTM! Left a note about some wording as a comment, but I don't think the wording *has* to be changed, it was just a suggestion.

systems/crusher_quick_start_guide.rst
verolero86: Thank you @hagertnl! I made some edits to make it clearer (hopefully).

PR #604, opened 2022-12-29
Update examples due to config changes on 12/29

verolero86:
This PR updates the Crusher Quick-Start examples with correct mappings that will be in effect after the outage on 12/29.


hagertnl (approved):
LGTM!

PR #616, opened 2023-01-24
Compress metrics to reduce  reruns in rocprof input file

hagertnl:
Compresses the metrics in the input.txt for roofline so that the application is re-run only twice (compared to 5x, previous).


Datawarrior85 (approved):
lgtm

PR #621, opened 2023-02-07
Added KI for CMake on Summit for HIP

hagertnl:
HIP code on Summit cannot be compiled when relying on HIP language support or hip::device and hip::host keywords. For HIP language support, CMake relies on using `hipconfig --rocm-path`, which on Summit, evaluates to `/opt/rocm`, which doesn't exist. `hip::device` and `hip::host` are defined in a CMake file in `lib/hip/cmake` that is not present in the `hip-cuda` installation, but is present in ordinary ROCm installations. So these keywords have no meaning on Summit, leading to compile failures.

### DIFF
--- a/systems/summit_user_guide.rst
+++ b/systems/summit_user_guide.rst
@@ -3737,10 +3737,14 @@ created with information on the files that were copied.
 Known Issues
 ============
 
-Last Updated: 30 December 2022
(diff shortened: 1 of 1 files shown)


Datawarrior85 (approved):
lgtm

PR #636, opened 2023-03-03
Added SBCAST section to Crusher & Frontier user guides

hagertnl:
Added preliminary SBCAST section for Crusher & Frontier, under the section heading "Tips for Launching at Scale". I can think of several other subsections that might be good under this heading, including some environment variables or discussion of directory structure within jobs (ie, don't make all ranks write to the same directory).


tom-papatheodore (approved):
This LGTM with the updates! Thanks @hagertnl!

PR #640, opened 2023-03-06
Added AQL library instructions to Frontier & Crusher user docs

hagertnl:
The AQL library is needed whenever using `rocprof` or `omniperf`. It is not detected by `ld`, since it is `dlopen`'d.
This library is opened regardless of whether any profiling is done or not, but the program will abort if you try to profile without it.


tom-papatheodore (approved):
LGTM

PR #664, opened 2023-04-05
Changed default sbcast to exclude some libraries on Crusher only

hagertnl:
Drafting a change in SBCAST instructions, which would leave the ROCm and CrayPE libraries in place, in their SquashFS file system. This change is shown to have negligible performance impact on Frontier.


tom-papatheodore (approved):
This LGTM, Nick!

PR #671, opened 2023-04-12
Re-added exclude=NONE section on Crusher & Frontier SBCAST

hagertnl:
Re-adding a paragraph & example script loaded with warnings for using `sbcast --exclude=NONE` on Frontier & Crusher.

### DIFF
--- a/systems/crusher_quick_start_guide.rst
+++ b/systems/crusher_quick_start_guide.rst
@@ -1249,68 +1249,6 @@ and here is the output from that script:
 
 ``sbcast`` also handles binaries and their libraries:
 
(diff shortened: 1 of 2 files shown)


tom-papatheodore (approved):
This LGTM. My only suggestion would be to clarify that the additional content is an _alternative_ to what was already shown above. Your text says it already, but if a user is scanning through the doc, that might not be clear. If you agree with this suggestion, I think you could just make the new content a sub-section called "alternate method" or something.

But again, this is just a suggestion. The PR is approved regardless. Nice (quick) work, my friend.

PR #672, opened 2023-04-13
Reformatting the Slurm introduction on Frontier.

LoganWGillum:
First contribution to OLCF User Docs


systems/frontier_user_guide.rst
```diff
@@ -14,7 +14,7 @@ Frontier User Guide
   Orion scratch filesystem 
     Frontier mounts Orion, a parallel filesystem based on Lustre and HPE ClusterStor, with a 679 PB usable namespace. Frontier will not mount Alpine and Summit will not mount Orion. Data will not be automatically transferred from Alpine to Orion, so we recommend that users move only needed data between the file systems with Globus.
```
michael-sandoval, line 15: Good catch on the below indentation, but for consistency I would also unindent this one by 1 space character so that it matches the other sections as well

michael-sandoval (changes requested):
Threw in a comment about indentation for the header but otherwise looks good to me!

michael-sandoval (approved):

PR #687, opened 2023-05-17
Linking to XPMEM with hipcc for Frontier and Crusher.

twhite-cray:
MPI needs to link with XPMEM for faster on-node MPI with host buffers. This update documents the necessary arguments to get XPMEM when linking with hipcc. (The HPE Cray compiler wrappers link with XPMEM automatically.)


verolero86 (approved):
thank you @twhite-cray !

PR #691, opened 2023-05-25
Modify reverse connect when using forge with local client

Datawarrior85:
Had to revert this back as the env vars need to be sourced on the compute side as well. The remote launch script with the vars is only needed if running forge client on local laptop


kellenrl (approved):
LGTM

PR #704, opened 2023-06-23
Fix HSA_XNACK=0 hipMallocManaged access behavior

abbotts:
When we determined that hipMallocManaged memory was backed by DDR with XNACK off we switched the "initial physical location" but forgot to update the behavior to make it consistent.

The table should now be correct for the current behavior.


tom-papatheodore (approved):
Thanks @abbotts! Approved and merged.

PR #705, opened 2023-06-23
Update andes_user_guide.rst

BranTran:
Double delimiter used in table causing delimiter to appear


michael-sandoval (approved):
Thanks for catching this! I expanded on your fix by switching the syntax to what the rest of our guides do (used `` syntax)

PR #707, opened 2023-06-28
Suzanne training

suzannepk:
Adding HPE slides and recordings now that we have a good legal agreement to setup to allow us to display them.


michael-sandoval (approved):
LGTM

PR #712, opened 2023-07-17
Adding documentation regarding BLAS usage of atomic operations for Frontier

daineAMD:
As stated. Using ROCm 5.5 and later, rocBLAS and hipBLAS introduced some atomic operations to increase performance. Users can turn this off in rocBLAS/hipBLAS as described if repeatable results are desired.

### DIFF
--- a/systems/frontier_user_guide.rst
+++ b/systems/frontier_user_guide.rst
@@ -3424,7 +3424,8 @@ Library considerations with atomic operations
 ---------------------------------------------
 
 Some functionality provided by the rocBLAS and hipBLAS libraries use atomic operations to improve performance by default. This can cause results to not be bit-wise reproducible.
-Functions that may use atomic operations include: gemv, hemv, and symv. These functions introduced atomic operations in ROCm 5.5. Under some circumstances gemm, gemm_ex, trsm, and trsv may also use atomic operations.
+Level 2 functions that may use atomic operations include: gemv, hemv, and symv, which introduced atomic operations in ROCm 5.5. All of the Level 3 functions, along with Level 2 trsv, may use atomic operations where dependent
+on gemm. Atomic operations are used for problem sizes where they are shown to improve performance.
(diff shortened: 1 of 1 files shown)


afanfa:
@tom-papatheodore

systems/frontier_user_guide.rst
```diff
@@ -3420,6 +3420,22 @@ The aforementioned behavior is only true for FP atomicAdd() operations. Hardware
 In cases when contention is very low, a FP32 CAS loop implementing an atomicAdd() operation could be faster than an hardware FP32 LDS atomicAdd().
 Applications using single precision FP atomicAdd() are encouraged to experiment with the use of double precision to evaluate the trade-off between high atomicAdd() performance vs. potential lower occupancy due to higher LDS usage.
 
+Library considerations with atomic operations
+---------------------------------------------
+
+Some functionality provided by the rocBLAS and hipBLAS libraries use atomic operations to improve performance by default. This can cause results to not be bit-wise reproducible.
+Functions that may use atomic operations include: gemv, hemv, and symv. These functions introduced atomic operations in ROCm 5.5. Under some circumstances gemm, gemm_ex, trsm, and trsv may also use atomic operations.
```
TorreZuk, line 3427: I think this could be a worded a bit clearer, whether they are used is based on problem sizes vs. "under some circumstances".

amcamd: The block recursive gemm based L3 functions call gemm, so they can also use atomic operations. I think it is correct to say all L3 functions may use atomic operations for some problem sizes.

amcamd: The above paragraph could be something like:
Level 2 functions that may use atomic operations include: gemv, hemv, and symv. These functions introduced atomic operations in ROCm 5.5. All the Level 3 functions may use atomic operations. The atomic operations are used for problem sizes where it improves performance.

tom-papatheodore:
@hagertnl will take a look at this since he's working most closely with these changes. Thanks for putting this together!

hagertnl (approved):
This section looks good to me. Thanks for putting this together!

PR #728, opened 2023-09-03
Update frontier scheduling policy

verolero86:


mattaezell (approved):
This matches the current settings.

PR #749, opened 2023-11-17
Update login node names for CITADEL

gounley:
Currently login node names are not active; this PR would replace them with current logins.


michael-sandoval (approved):
LGTM

PR #756, opened 2023-12-21
Suzanne endpoints

suzannepk:
I have fixed all the data transfer docs to mention the Globus 5  endpoint and I have commented Alpine out of the data guides.


michael-sandoval (changes requested):
Andes guide also needs to change because of Alpine references. (around line 85). The andes guide now produces an error pointing to a removed reference from the changes

michael-sandoval:
Also, forgot to mention that this probably needs to be changed as well in the data section:

> These directories should be used for storage of data not immediately needed in either the Project Home (NFS) areas or Project Work (Alpine) areas and to serve as a location to store backup copies of project-related files.

From Alpine to Orion (line 293)

michael-sandoval (approved):
LGTM

PR #761, opened 2024-01-05
Verolero86 ace

todbadrakh:
Added Defiant quick-start guide to ACE Testbed with the associated changes in index.rst links:
- new node diagram
- changed to correct amd compiler paths and modules
- fixed amd openmp offload flag example (--target)

Need info on:
- file system and directory structure
- slurm queues (batch-cpu and batch-gpu) and their policies


ace_testbed/defiant_quick_start_guide.rst
```diff
@@ -0,0 +1,1142 @@
...
+1. Compiling with the Cray compiler wrappers, ``cc`` or ``CC``
+""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+
+To use GPU-aware Cray MPICH with the Cray compiler wrappers, users must load
+specific modules, set some environment variables, and include appropriate headers
+and libraries. The following modules and environment variables must be set:
+
+.. note:: 
+
+    Setting ``MPICH_SMP_SINGLE_COPY_MODE=CMA`` is required as a temporary
```
secondspass, line 279: This is referencing a known issue that seems to be resolved. It might've been something that was present in spock. Should verify its still an issue.

ace_testbed/defiant_quick_start_guide.rst
```diff
@@ -0,0 +1,1142 @@
...
+    issue has been resolved.
+
+.. code:: bash
+
+    module load craype-accel-amd-gfx908
+    module load PrgEnv-cray
+    module load amd
+
+    ## These must be set before running
+    export MPIR_CVAR_GPU_EAGER_DEVICE_MEM=0
```
secondspass, line 292: Are these env variables still necessary to use GPU aware MPI? I'm not 100% sure that GPU aware MPI even works correctly on Defiant yet when I tried, but I might be doing it wrong. That's something to verify.

index.rst
```diff
@@ -46,6 +46,7 @@ Systems
    :maxdepth: 2
 
    systems/index
+   ace_testbed/index
```
secondspass, line 49: this one and the index item in line 106 below is causing the link to the ACE testbed doc to appear twice on the sidebar. There should only be one (though I don't know which one would be better). My vote is to keep the one on line 106 and delete this one.

secondspass (commented):
Hey Togo, this is Subil. Thank you for working on the docs. I've left a few comments to check after going through it. Let me know if I can be of any help.

ace_testbed/defiant_quick_start_guide.rst
todbadrakh: Issue is resolved and we are at MPICH version 8.1.27.

index.rst
todbadrakh: Forgot to delete this one at line 49 after adding line 106. Removing line 49.

ace_testbed/defiant_quick_start_guide.rst
todbadrakh: Leaving for now until confirmed.

PR #768, opened 2024-01-23
Update Frontier guide with latest software versions

verolero86:
Adds information about the newly installed PE and ROCm modulefiles.


hagertnl (approved):
LGTM!

PR #773, opened 2024-01-25
Update index.rst

suzannepk:
Adding in specs for Alpine2


suzannepk:
This change should not break it since it was only text, but I have not rendered the site.

michael-sandoval (approved):
LGTM

PR #781, opened 2024-02-01
Updated MPICH & ROCm compatibility table, removed amdgpu-target mentions

hagertnl:
This PR is updating MPICH & ROCm compatibility and clarifying what we mean by compatibility. This includes up to CrayPE 23.12.

Also noticed that amdgpu-target is still used, so I replaced that with offload-arch.


michael-sandoval (approved):
LGTM

PR #782, opened 2024-02-01
WIP: Updated docs for summit plus containers

AcerP-py:
Updated docs for Summit Plus and the new olcf base images.


secondspass (changes requested):
Hey Asa! Great job with updating the docs. Here are my comments:
1. Change all references to Singularity to Apptainer since we're using Apptainer now. 
2. Around line 380 where you have the submit.lsf for the single node gpu example, add `module unload spectrum-mpi` to unload the mpi module, to prevent the error message coming up for libpami_cudahook.so not being visible inside the container (when it is not even needed for a non MPI example).

secondspass (approved):

PR #796, opened 2024-03-05
Added software page for PyTorch

AcerP-py:
Add page to docs for PyTorch.

### DIFF
--- a/software/python/index.rst
+++ b/software/python/index.rst
@@ -56,14 +56,14 @@ Below is a list of guides created for using Python on OLCF systems.
    cupy
    sbcast_conda
    jupyter_envs
-   pytorch
+   pytorch_summit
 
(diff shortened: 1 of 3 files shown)


michael-sandoval (changes requested):
Just for clarification, might be good to indicate how a user can tell which Python version a wheel will work with (e.g., pointing out how `cp310` or `cp311` mean to use Python 3.10 or 3.11) -- or you can create a compatibility table that indicates which wheel works with which version.

michael-sandoval:
Oh, also, it may be worth briefly pointing out in the "intro" (or as a note somewhere) that using the `open-ce` module is also an option but is a more complex environment than just PyTorch. And/or mentioning that your guide is for those that don't need everything in OpenCE and just want Pytorch. And then linking to the OpenCE page for users interested in using that instead.

AcerP-py:
Ok, I think I got both of those done. I also added a few things I missed before and swapped the order of a few items.

PR #817, opened 2024-04-17
Fixing KDI vs SPI differences

LoganWGillum:
Chris informed me that the login process for Frontier SPI is different from KDI vs whitelisted SPI.


michael-sandoval (approved):
LGTM

PR #833, opened 2024-05-30
Add basic instructions for using SYCL on Frontier

npmiller:
This patch adds basic instructions to use the oneAPI modules to use SYCL on Frontier.


michael-sandoval (approved):
LGTM. Docs build fine, and module loads / compilers work

PR #846, opened 2024-07-10
Added message about upcoming module layout changes

AcerP-py:


software/software-news.rst
```diff
@@ -26,6 +26,13 @@ As a result of the upgrade, the following packages are now default:
     "GNU Compiler Collection (GCC)", "12.2.0", "12.3.0"
     "AMD Compiler (AMD)", "5.3.0", "5.7.1"
 
+In addition to the default modules changing, the layout of the modules for the OLCF provided software stack will change.
+Going forward our core software (i.e software independent of compiler type/version) will be in a default module
+called `Core` with the version being in the format `year/month`. Additionally the suffixes to modules have
```
hagertnl, line 31: Maybe instead of "Additionally the suffixes to modules have been reworked", we can say "Explicit suffixes have been added to modules to indicate MPI, OpenMP, and GPU support." Technically, we had very few suffixes before, so we didn't rework too much, right?

software/software-news.rst
```diff
@@ -26,6 +26,13 @@ As a result of the upgrade, the following packages are now default:
...
+been reworked. All modules that depend on ROCm are suffixed with `-gpu`. The same is true for packages with mpi `-mpi` and
+openmp `-omp`. Software is built to target CPE releases based on compiler (e.g. cce, amd or gcc), cray-mpich, and ROCm versions.
+If you have an unsupported combination of those modules you may not see certain parts of the normal software stack.
```
hagertnl, line 34: Maybe add a reference to https://docs.olcf.ornl.gov/systems/frontier_user_guide.html#determining-the-compatibility-of-cray-mpich-and-rocm regarding "unsupported"?

software/software-news.rst
AcerP-py: I agree. Should be updated now.

verolero86 (approved):

software/software-news.rst
AcerP-py: That line is actually misleading. We are not actually building for every supported combination. Maybe rephrase to "We support CPE releases"

AcerP-py: :+1:

PR #847, opened 2024-07-10
updating Quantinuum docs

AstroLando:
Updating wording to remove mention of max qubit count given QTM's increasing upgrade frequency


michael-sandoval (approved):
LGTM

PR #855, opened 2024-07-19
Removed ROCm/5.1.0, 5.2.0 from MPICH/ROCM compatibility table

hagertnl:
Removed ROCm/5.1.0, 5.2.0 from the user docs. This is consistent with what we did with ROCm/4 -- we only list the ROCm versions we currently provide in this table.


ddietz89 (approved):
LGTM

PR #894, opened 2024-11-28
Updating IBMQ queue policy and removing mentions of reservations

AstroLando:
Updating IBMQ queue policy to reflect 100m default limit and removed mentions of reservations


michael-sandoval (approved):
LGTM

PR #898, opened 2024-12-04
Removing Rigetti references from Quantum Docs

AstroLando:
Removed Rigetti references and content from Quantum docs as we no longer work with Rigetti.  Removed Rigetti guide, removed hello qcup scripts, and descriptions from access guide.


michael-sandoval (approved):
LGTM

PR #902, opened 2024-12-19
Removed module spider warning, added content to the ROCm/MPICH compatibility sections

hagertnl:
Removed the module spider warning, since we now have a new module spider utility on Frontier which contains patches that fix the earlier issues.

I also took this opportunity to improve the ROCm/MPICH compatibility sections, including adding a new note that describes how to use the CrayPE `rc.lua` files to effectively set your own default modules.


michael-sandoval (approved):
LGTM

PR #915, opened 2025-01-24
Jordan edits branch2

jw676:
Mostly formatting changes and link fixes/additions for better navigation. There were a few references to Summit that I removed. In particular, the "Software News" page had references to decommissioned systems (i.e., Summit and Rhea) that I deleted, but I'm not sure if they should be kept for bookkeeping purposes.

### DIFF
--- a/software/software-news.rst
+++ b/software/software-news.rst
@@ -8,7 +8,7 @@ most recent changes are listed first.
 
 ----
 
-Frontier: Updated modules for cpe/23.12 (October 16 2024)
+Frontier: Updated Modules for cpe/23.12 (October 16 2024)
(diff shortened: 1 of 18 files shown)


software/software-news.rst
secondspass, line 1: Lets avoid removing information about old systems from this page till we can decide they should be removed. This does have a lot of historical information and I think its useful to keep that here.

secondspass (changes requested):

jw676:
Summit + Rhea sections have been readded to the software news page, so now the branch is ready to go.

secondspass (approved):

PR #929, opened 2025-02-18
Updated to make verbs past-tense for February 18 outage

hagertnl:
Also triggers a re-build of the Known Issues list, which has been updated.


michael-sandoval (approved):
LGTM

PR #943, opened 2025-03-20
Add software catalog page

ayarkhan:
This needs a software_list_group.json file to be placed in software/software_list by the documentation backend infrastructure.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -9,6 +9,8 @@ http://www.sphinx-doc.org/en/master/config
 # pylint: disable=import-error, invalid-name, redefined-builtin
 
 import datetime as dt
+import os
+import shutil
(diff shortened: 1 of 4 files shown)


hagertnl:
I converted this to draft just to make sure someone doesn't accidentally merge into the public docs yet. This needs to be tested on the internal mirror first, since there's OpenShift-side functionality that has to be tested as well.

@ayarkhan this looks great! Can you change the description/homepage to check for the string "None" instead of being null? Also, I believe that every software should now have the nameVersion field after the latest fix to our query script, so the "nan" check should be superfluous if you would like to remove it.

kellenrl:
looks good to me

kellenrl (approved):

PR #949, opened 2025-04-14
Updated tau docs to match new tau modules

ayarkhan:


software/profiling/TAU.rst
```diff
@@ -214,7 +217,7 @@ Profiling and tracing for code execution follow the earlier example.
 
   # Allocate 1 node for 5 min and run 2 tasks 
   # Note: This is not using tau_exec
-  srun -A <ACCOUNT> -N 1 -t 5 -n 2 ./computePi_taucxx
+  srun -A $SBATCH_ACCOUNT -N 1 -t 5 -n 2 ./computePi_taucxx
```
Datawarrior85, line 220: Some confusion with using $SBATCH_ACCOUNT. This would require the user to mod their bashrc or set the var in the environment. Chatted offline with Asim and will update to use < > placeholder switch .

ayarkhan: Done

Datawarrior85 (approved):

Datawarrior85 (approved):

PR #968, opened 2025-06-04
Updating Industry Reporting Templates

depetromg:
- Updated the existing industry quarterly report template with the current version
- Added the industry closeout report template
- These will be accessed by users who receive reporting notification emails via RATS-CRM


michael-sandoval (approved):
LGTM

PR #977, opened 2025-07-14
Updates to Jupyter docs (Slate).

jw676:


michael-sandoval (changes requested):
Would it be possible to scale down the `jupyterlab_images2.png` picture (in the page itself, not the raw image) to about 75% of how big it is on the page now? If not I think it's all fine, but that's the only thing that stuck out to me

PR #978, opened 2025-07-15
Updated VisIt and ParaView guides

michael-sandoval:
added pv host profiles to repo

updated visit host profile on page

### DIFF
--- a/software/viz_tools/paraview.rst
+++ b/software/viz_tools/paraview.rst
@@ -146,9 +146,9 @@ methods may be used, the one described should work in most cases.
 
 **Step 1: Save the following PVSC files to your local computer**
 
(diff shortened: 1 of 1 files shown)


jw676:
Downloadable pvsc files is a nice change. However, I was testing with Paraview 5.13.1 and I cannot connect to Frontier using either of the files (Andes is fine). I'm not sure if it's a specific issue with my setup or something we should investigate further before pushing through. I'll approve after I can get a sanity check.

michael-sandoval:
Logan just tested with 5.13.1 / 5.13.2 and was able to get in with the new host profiles, so could be your setup @jw676

jw676 (approved):

PR #990, opened 2025-08-27
Update hello qcup versions and instantiation.

t-ramz:


jw676 (approved):
Confirmed newer qiskit versions work

PR #993, opened 2025-09-02
updating Andes paraview versions to include 6.0.0

warenaud:
Updating to add newly installed versions of Paraview


michael-sandoval:
LGTM (and paraview install seems to work)

michael-sandoval (approved):

PR #996, opened 2025-09-05
Fix documentation for using sbcast for ROCm 6.4.x binaries

timattox:
With the newer ROCm 6.4.x series, there is a new dlopen issue when using the "SBCASTing a binary with ALL libraries" technique for Frontier at scale. Added in the lines to the example script to add synbolic links for libamd_comgr.so


hagertnl (approved):

hagertnl:
Thanks for authoring this, @timattox !

PR #1016, opened 2025-10-24
First pass at updating prose to match user guide

ddietz89:


software/workflows/rp.rst
```diff
@@ -105,42 +82,43 @@ batch system for a submission (this is pulled from the predefined configuration)
...
+------
 
-In order to run the following you should first
+The repository for this example can be reached here: https://github.com/henrymonge/chroma_rp.
 
 .. code-block:: console
 
     $ git clone https://github.com/henrymonge/chroma_rp.git
     $ cd chroma_rp
+    $ ./compile_chroma/build_stack.sh
```
t-ramz: neat I imagine this is why I ran into errors that were hard to debug

ddietz89: Will add a note to edit the env.sh as there's some module loads and what not that may not work. Looks like you can set install paths and such in there too.

michael-sandoval:
Going through old PRs to clean up things. @t-ramz and @ddietz89 is this still being worked on?

PR #1023, opened 2025-12-01
Add CPE 25.09 Containers

AcerP-py:

PR #1053, opened 2026-03-18
Top-level container example is for rocm/6.2.4 to match examples repo

markstock:
Since https://github.com/olcf/olcf_containers_examples/pull/9 was merged, this is the corresponding documentation update. It updates the file names and software versions for the default container example.

This should not conflict with a parallel PR: https://github.com/olcf/olcf-user-docs/pull/1023

### DIFF
--- a/software/containers_on_frontier.rst
+++ b/software/containers_on_frontier.rst
@@ -28,7 +28,7 @@ containers on Frontier. Some of the examples will refer to files in the
 Building and running a container image from a base Linux distribution for MPI
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-This example will walk through building up a container image with OpenSUSE 15.4, MPICH 3.4.2, and ROCm 5.7.1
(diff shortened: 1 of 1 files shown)


software/containers_on_frontier.rst
```diff
@@ -28,7 +28,7 @@ containers on Frontier. Some of the examples will refer to files in the
 Building and running a container image from a base Linux distribution for MPI
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-This example will walk through building up a container image with OpenSUSE 15.4, MPICH 3.4.2, and ROCm 5.7.1
+This example will walk through building up a container image with OpenSUSE 15.6, MPICH 3.4a2, and ROCm 6.2.4
```
t-ramz: I'm not sure it makes an a big difference, but the actual version of MPICH in the container is 3.4.2, which I believe is now accurate to leave alone in docs.

t-ramz: I see that is a change more recent than this PR!
I'll fix and merge :)

t-ramz (approved):

PR #1064, opened 2026-04-02
Slurm 44 updates

t-ramz:
I took changes provide by AJ and re-organized them with tabs per section for easier naviagtion


michael-sandoval (approved):
LGTM and confirmed with the S3M folks that this is accurate

PR #1069, opened 2026-04-13
CuPy and Jax: updated installation guides to ROCm 7

michael-sandoval:


software/analytics/jax.rst
```diff
@@ -49,7 +54,7 @@ Loading a python module puts you in a "base" environment, but you need to create
 
       .. code-block:: bash
 
-         conda create -n jax_env-frontier python=3.11 numpy scipy -c conda-forge
+         conda create -n jax_env-frontier python=3.12 numpy scipy -c conda-forge
```
t-ramz: I think now would be a good opportunity to make these environment creations match the path specification instead. That makes it more portable to Orion, at least.

michael-sandoval: done

t-ramz (approved):
lgtm

PR #1072, opened 2026-04-27
Naming Conventions Note

jilamni:
made Security Policy section more clear and added note about naming conventions in Summary of Storage Areas


t-ramz (approved):
lgtm

data/index.rst
```diff
@@ -23,6 +23,15 @@ Each user has a User Home area on NFS. Each project has a Project Home area on N
 - **Project Archive:** Long-term project data for archival access that's shared with other project members.
 - **World Archive:** Long-term project data for archival access that's shared with users outside your project.
 
+.. warning::
```
t-ramz, line 26: Not relevant to this PR but I think it would be cool to have a standardized way for our docs where we extract templates from repeated text we have in multiple places. That way you could insert this as a template instead.

PR #1075, opened 2026-05-15
Inference Service: fixed syntax error

michael-sandoval:


services_and_applications/olcf_inference/index.rst
```diff
@@ -529,7 +529,7 @@ To view a list of available models use the ``/models`` endpoint.
 
 **Expected JSON Response:**
 
-.. code-block:: json
+.. code-block::
```
t-ramz: `json` is actually correct, I'm pretty sure it's just missing a trailing `]` and `}` as to why it's being highlighted

t-ramz (approved):

PR #1081, opened 2026-05-20
Add details on IPC cache

ashesh2512:
This PR adds information on the combination of environment variables required to enable stable operation of IPC. Added best practices. Added a summary on known issues associated with IPC cache, and related workarounds. Also added a description on relevant environment variables.


abbotts:
@trey-ornl , @GeorgiadouAntigoni , @hagertnl - I'd like to talk about this at our next GPU MPI meeting (I'll forward the invite to @ashesh2512 ). 

The hope is that in the next MPI + ROCM release all the edge case memory leaks will be resolved and we can just strop back to the "here's how to enable the IPC signal cache and tune the IPC cache size" documentation.

systems/frontier_user_guide.rst
```diff
@@ -4059,6 +4079,30 @@ Using GPU-aware MPI is highly recommended on Frontier because the HPE Slingshot
...
+
+``GTL_ENABLE_HSA_IPC_SIGNAL_CACHE``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This option is unset by default, and is advised to be set to ``GTL_ENABLE_HSA_IPC_SIGNAL_CACHE=1`` when ``MPICH_GPU_IPC_ENABLED=1``. 
+
+Registering and unregistering GPU buffers for inter-process communication takes a relatively long time to do. The MPI library caches remote buffer attachments in order to reuse them, instead of detaching after every transfer. One limitation has been that when a process stops using a GPU buffer, there was no easy way to let the processes that are caching attachments know about this, and so they continued to cache the buffer. This consumes some of the limited GPU memory. A new feature since ROCM 6.4 enables the MPI library to alert the other processes through a signalling mechanism. If this environment variable is set, then this feature will be used to help reduce the amount of memory that the cache is consuming. This may help prevent out of memory situations. When enabling this feature, we do not recommend using the memhooks ``FI_MR_CACHE_MONITOR``, as it has been shown to hang in certain situations.
```
abbotts: I think you can remove the memhooks sentence here. We don't recomment memhooks with GPU aware MPI at *any* point.

systems/frontier_user_guide.rst
```diff
@@ -4059,6 +4079,30 @@ Using GPU-aware MPI is highly recommended on Frontier because the HPE Slingshot
...
+
+Registering and unregistering GPU buffers for inter-process communication takes a relatively long time to do. The MPI library caches remote buffer attachments in order to reuse them, instead of detaching after every transfer. One limitation has been that when a process stops using a GPU buffer, there was no easy way to let the processes that are caching attachments know about this, and so they continued to cache the buffer. This consumes some of the limited GPU memory. A new feature since ROCM 6.4 enables the MPI library to alert the other processes through a signalling mechanism. If this environment variable is set, then this feature will be used to help reduce the amount of memory that the cache is consuming. This may help prevent out of memory situations. When enabling this feature, we do not recommend using the memhooks ``FI_MR_CACHE_MONITOR``, as it has been shown to hang in certain situations. 
+
+``HSA_ENABLE_IPC_MODE_LEGACY``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This specifies AMD's IPC implementation, and should be set to ``1`` when ``MPICH_GPU_IPC_ENABLED=1``. Cray MPICH 9.0.1 and before should use ``HSA_DISABLE_IPC_MODE_LEGACY=0``.
```
abbotts: I think this should note that HSA_ENABLE_IPC_LEGACY_MODE=1 is currently default on Frontier. I'm not sure about the cray-mpich/9.0.1 guidance. Why do you recommend users should change it for 9.0.1 instead of just using the default?

systems/frontier_user_guide.rst
```diff
@@ -670,10 +670,15 @@ To use GPU-aware Cray MPICH with Frontier's PrgEnv modules, users must set the f
     module load craype-accel-amd-gfx90a
     module load rocm
     
-    export MPICH_GPU_SUPPORT_ENABLED=1    
-
+    export MPICH_GPU_SUPPORT_ENABLED=1
+    export MPICH_GPU_IPC_ENABLED=1 # enabled by default if MPICH_GPU_SUPPORT_ENABLED=1
```
trey-ornl: If it is default, why do users need to add it?

systems/frontier_user_guide.rst
```diff
@@ -4015,6 +4020,21 @@ If it is necessary to have bit-wise reproducible results from these libraries, i
...
+
+Another enhancement in GPU-Aware MPI that relates to IPC, has to do with multiple messages in flight. Prior to Cray MPICH 8.1.33, if the maximum messages in flight at any given time exceeded ``MPICH_GPU_IPC_CACHE_MAX_SIZE``, the program would abort immediately. This has been resolved, and IPC cache handles are evicted as the number of messages in flight increase beyond ``MPICH_GPU_IPC_CACHE_MAX_SIZE``. However, a new issue exists where memory leaks can occur if GPU buffers are repeatedly created and destroyed while the number of messages in flight exceed ``MPICH_GPU_IPC_CACHE_MAX_SIZE``. To mitigate this, it is recommended to set ``MPICH_GPU_IPC_CACHE_MAX_SIZE`` to a value larger than the maximum anticipated messages in flight. The default value is ``50``. If setting ``MPICH_GPU_IPC_CACHE_MAX_SIZE > 50``, the following parameters should be adjusted accordingly to maintain a decreasing order of magnitude, i.e., ``GTL_MAX_SIG_HANDLER_LEN`` >  ``GTL_IPC_SIGNAL_POOL_TABLE_LEN`` >= ``MPICH_GPU_IPC_CACHE_MAX_SIZE``.
```
trey-ornl: At the very end, the inequality is not an example but a requirement. I suggest changing the ", i.e.," to ":". Further, I suggest adding, "For example:", with some actual working settings for these variables.

systems/frontier_user_guide.rst
```diff
@@ -4059,6 +4079,30 @@ Using GPU-aware MPI is highly recommended on Frontier because the HPE Slingshot
...
+
+``GTL_ENABLE_HSA_IPC_SIGNAL_CACHE``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This option is unset by default, and is advised to be set to ``GTL_ENABLE_HSA_IPC_SIGNAL_CACHE=1`` when ``MPICH_GPU_IPC_ENABLED=1``. 
+
+Registering and unregistering GPU buffers for inter-process communication takes a relatively long time to do. The MPI library caches remote buffer attachments in order to reuse them, instead of detaching after every transfer. One limitation has been that when a process stops using a GPU buffer, there was no easy way to let the processes that are caching attachments know about this, and so they continued to cache the buffer. This consumes some of the limited GPU memory. A new feature since ROCM 6.4 enables the MPI library to alert the other processes through a signalling mechanism. If this environment variable is set, then this feature will be used to help reduce the amount of memory that the cache is consuming. This may help prevent out of memory situations.
+
+``HSA_ENABLE_IPC_MODE_LEGACY``
```
trey-ornl, line 4092: I think you need to document both versions of this variable, including the Rocm version dependence.

systems/frontier_user_guide.rst
```diff
@@ -4015,6 +4020,21 @@ If it is necessary to have bit-wise reproducible results from these libraries, i
...
 ------
 
+.. _gpu-ipc-tips-tricks:
+
+GPU inter-process communication (IPC) considerations
+----------------------------------------------------
+
+Cray MPICH uses IPC by default for intra-node inter-process MPI data movement operations that involve GPU buffers, i.e., enabling GPU-aware MPI (``MPICH_GPU_SUPPORT_ENABLED=1``) also sets ``MPICH_GPU_IPC_ENABLED=1``. Disabling IPC is known to noticeably impact intra-node MPI performance.
+
+Prior to Cray MPICH 8.1.33, a memory leak was observed in edge cases involving repeated creation and freeing of GPU-buffers used in MPI communication. This issue was prevelant when GPU buffers continuously increased in size or occupied new memory regions (e.g., due to defragmentation). The root cause was traced to IPC cache handles not being properly released. AMD addressed this by introducing an API that allows remote processes to free IPC cache handles. For Cray MPICH 9.0.1 and prior, the updated IPC mechanism is enabled by setting ``GTL_ENABLE_HSA_IPC_SIGNAL_CACHE=1`` and ``HSA_DISABLE_IPC_MODE_LEGACY=0``. Starting with Cray MPICH 9.1.0, the latter variable was renamed to ``HSA_ENABLE_IPC_MODE_LEGACY=1`` for improved clarity.
```
trey-ornl: Doesn't HSA_ENABLE/DISABLE_IPC_MODE_LEGACY depend on Rocm version, not Cray MPI version?

systems/frontier_user_guide.rst
```diff
@@ -4015,6 +4020,21 @@ If it is necessary to have bit-wise reproducible results from these libraries, i
...
+
+A separate issue remains where enabling IPC can still lead to memory leaks for message sizes less than or equal to 1 MB. A workaround is to set ``HSA_ENABLE_IPC_MODE_LEGACY=0`` and ``GTL_HSA_VSMSG_CUTOFF_SIZE=0``. When message sizes fall at or below ``GTL_HSA_VSMSG_CUTOFF_SIZE``, the GPU Transport Layer (GTL) uses a blocking memory copy driven by the CPU instead of an HSA memcpy. For context, disabling legacy IPC mode (``HSA_ENABLE_IPC_MODE_LEGACY=0``) switches to AMD's newer IPC implementation based on the Linux dma_buf mechanism. 
+
```
trey-ornl, line 4031: Is it true that HSA_ENABLE_IPC_MODE_LEGACY=0 is only supported by Cray MPI iff GTL_HSA_VSMSG_CUTOFF_SIZE=0?

systems/frontier_user_guide.rst
```diff
@@ -4015,6 +4020,21 @@ If it is necessary to have bit-wise reproducible results from these libraries, i
...
+A separate issue remains where enabling IPC can still lead to memory leaks for message sizes less than or equal to 1 MB. A workaround is to set ``HSA_ENABLE_IPC_MODE_LEGACY=0`` and ``GTL_HSA_VSMSG_CUTOFF_SIZE=0``. When message sizes fall at or below ``GTL_HSA_VSMSG_CUTOFF_SIZE``, the GPU Transport Layer (GTL) uses a blocking memory copy driven by the CPU instead of an HSA memcpy. For context, disabling legacy IPC mode (``HSA_ENABLE_IPC_MODE_LEGACY=0``) switches to AMD's newer IPC implementation based on the Linux dma_buf mechanism.
```
trey-ornl, line 4030: I think you might need to include a reminder about the older HSA_DISABLE_IPC_MODE_LEGACY=1 version too.

systems/frontier_user_guide.rst
ashesh2512: The users don't need to add it to use IPC, but documenting the variable is helpful. In the past toggling this variable has helped teams identify and report bugs with IPC.

trey-ornl: I vote for removing it here because it is documented below, but I don't mind being outvoted. Maybe add below that turning it off can be helpful for debugging IPC?

ashesh2512: Makes sense. I have removed it here.

systems/frontier_user_guide.rst
ashesh2512: Added example.

systems/frontier_user_guide.rst
ashesh2512: Thanks for catching that. I have added the older version.

systems/frontier_user_guide.rst
trey-ornl: Aha! Yes, it's RocR, not Cray MPICH.
https://github.com/ROCm/rocm-systems/blame/bece9cc90b397fc420ffe699ee812298a95e99db/projects/rocr-runtime/runtime/hsa-runtime/core/util/flag.h#L265

trey-ornl: I can find no evidence of HSA_DISABLE_IPC_MODE_LEGACY in any RocR release. It looks like HSA_ENABLE_IPC_MODE_LEGACY appeared in tag rocm-6.1.0, with no option, neither DISABLE or ENABLE, in the previous tag, rocm-6.0.2.
https://github.com/ROCm/ROCR-Runtime/blob/a4ab29c549573e8f02067cdd2a8e21f82d1a6958/src/core/util/flag.h#L228
Are you sure the DISABLE version exists anywhere on Frontier? Maybe it was only in a debug version HPE got from AMD, which they don't appear to have made public?

ashesh2512: You are right in that there is no `HSA_DISABLE_IPC_MODE_LEGACY`. I have unfortunately mistyped that throughout the PR when I was trying to distinguish between  `GTL_ENABLE_HSA_IPC_SIGNAL_CACHE` and `GTL_DISABLE_HSA_IPC_SIGNAL_CACHE`. I need to clean that up.

trey-ornl: I'm just glad I never make mitsakes.

ashesh2512:
@trey-ornl Could you please give this another look? I have removed the incorrect environment variable from the PR.

systems/frontier_user_guide.rst
trey-ornl: Github won't let me resolve this comment.

ashesh2512: I am able to

systems/frontier_user_guide.rst
```diff
@@ -4059,6 +4078,30 @@ Using GPU-aware MPI is highly recommended on Frontier because the HPE Slingshot
 There is a small but measurable latency impact for enabling GPU-aware MPI with CPU buffers.
 Applications that do not use GPU buffers in MPI calls may want to leave this variable unset or set to ``0``, especially if they are sensitive to small message latency.
 
+``MPICH_GPU_IPC_ENABLED``
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Set to ``1`` by default if ``MPICH_GPU_SUPPORT_ENABLED=1``. This enables GPU IPC support for intra-node GPU-GPU communication operations. This variable has no effect if MPICH_GPU_SUPPORT_ENABLED is set to 0. Disabling IPC is known to noticeably impact intra-node MPI performance. For known issues with IPC enabled and suggested workarounds, please see :ref:`gpu-ipc-tips-tricks`.
+
+``GTL_ENABLE_HSA_IPC_SIGNAL_CACHE``
```
trey-ornl: Add to the title: "/ ``GTL_DISABLE_HSA_IPC_SIGNAL_CACHE``"?

systems/frontier_user_guide.rst
```diff
@@ -4059,6 +4078,30 @@ Using GPU-aware MPI is highly recommended on Frontier because the HPE Slingshot
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This option is unset by default, and is advised to be set to ``GTL_ENABLE_HSA_IPC_SIGNAL_CACHE=1`` when ``MPICH_GPU_IPC_ENABLED=1``.
```
trey-ornl: Explain up front the name change between versions?

trey-ornl: What is the default behavior for each version of this environment variable when it is unset?

systems/frontier_user_guide.rst
ashesh2512: I had that initially, but decided against it. Adding `GTL_DISABLE_HSA_IPC_SIGNAL_CACHE` in the heading made it seam like the two variables were interchangeable which could be misconstrued.  If you feel it should definitely be there, happy to add it back in.

trey-ornl: How about this?
``GTL_ENABLE_HSA_IPC_SIGNAL_CACHE`` (previously ``GTL_DISABLE_HSA_IPC_SIGNAL_CACHE``)
I want to make sure that a search for either also shows the other one.

ashesh2512: that makes sense. I have made the change.

systems/frontier_user_guide.rst
```diff
@@ -670,10 +670,14 @@ To use GPU-aware Cray MPICH with Frontier's PrgEnv modules, users must set the f
     module load craype-accel-amd-gfx90a
     module load rocm
     
-    export MPICH_GPU_SUPPORT_ENABLED=1    
-
+    export MPICH_GPU_SUPPORT_ENABLED=1
+    export GTL_ENABLE_HSA_IPC_SIGNAL_CACHE=1 # Cray MPICH 9.1.0 onwards
+    export GTL_DIABLE_HSA_IPC_SIGNAL_CACHE=0 # Cray MPICH 9.0.1 and before
```
abbotts: Typo here - should be DISABLE not DIABLE

abbotts: I'm actually not sure if we want to have this suggestion here in the base 9.0.1 recipe to enable GPU aware MPI. I think I'd rather we just tell users "if you're using 9.1.0 or newer, then use the signal cache whenever you use GPU aware MPI." Then the detailed documentation below can show them how to enable it on an earlier version if they want. I just don't think we want everyone using 9.0.1 to enable the signal cache, because it was kind of broken there.

systems/frontier_user_guide.rst
```diff
@@ -670,10 +670,14 @@ To use GPU-aware Cray MPICH with Frontier's PrgEnv modules, users must set the f
...
+    export HSA_ENABLE_IPC_MODE_LEGACY=1
```
abbotts, line 675: This should be set on Frontier by default.

systems/frontier_user_guide.rst
```diff
@@ -4059,6 +4078,30 @@ Using GPU-aware MPI is highly recommended on Frontier because the HPE Slingshot
...
 
+``MPICH_GPU_IPC_ENABLED``
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Set to ``1`` by default if ``MPICH_GPU_SUPPORT_ENABLED=1``. This enables GPU IPC support for intra-node GPU-GPU communication operations. This variable has no effect if MPICH_GPU_SUPPORT_ENABLED is set to 0. Disabling IPC is known to noticeably impact intra-node MPI performance. For known issues with IPC enabled and suggested workarounds, please see :ref:`gpu-ipc-tips-tricks`.
+
+``GTL_ENABLE_HSA_IPC_SIGNAL_CACHE`` (previously ``GTL_DISABLE_HSA_IPC_SIGNAL_CACHE``)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This option is set to ``0`` by default, and is advised to be set to ``GTL_ENABLE_HSA_IPC_SIGNAL_CACHE=1`` when ``MPICH_GPU_IPC_ENABLED=1``. For Cray MPICH 9.0.1 and prior, set ``GTL_DISABLE_HSA_IPC_SIGNAL_CACHE=0`` which is set to ``1`` by default.
```
abbotts: This is _only_ 8.1.33 and 9.0.1, not 9.0.1 and prior.

systems/frontier_user_guide.rst
```diff
@@ -670,10 +670,14 @@ To use GPU-aware Cray MPICH with Frontier's PrgEnv modules, users must set the f
     module load craype-accel-amd-gfx90a
     module load rocm
     
-    export MPICH_GPU_SUPPORT_ENABLED=1    
-
+    export MPICH_GPU_SUPPORT_ENABLED=1
+    export GTL_ENABLE_HSA_IPC_SIGNAL_CACHE=1 # Cray MPICH 9.1.0 onwards
```
abbotts: I might add a comment here saying "Optional but recommended with Cray MPICH 9.1.0" or something to indicate that this isn't required but is recommended.

abbotts (commented):
I think we're almost there. I think we should tighten when we suggest users enable the new cache mechanism. I'm going to be a little conservative and say that we should _document_ how to enable it with 8.1.33 and 9.0.1 but not recommend it, since it was kind of broken there. We should keep the addition as-is to recommend it for 9.1.0 since it mostly works there.

systems/frontier_user_guide.rst
ashesh2512: I have changed this here and other places this was worded similarly

systems/frontier_user_guide.rst
ashesh2512: it is. the default value of the environment variable is stated when the variable is described in the documentation. This section is just stating the most relevant set of environment variables associated with GPU-aware MPI on Frontier. 

Although, since it is automatically set when `MPICH_GPU_SUPPORT_ENABLED=1`. I suppose I am leaning towards removing it here.

abbotts: just to be clear, it's not toggled on when MPICH_GPU_SUPPORT_ENABLED=1, it's just always set. I think it's fine to leave it here, though.

abbotts:
@michael-sandoval , @hagertnl  , @trey-ornl , @ashesh2512  - it would be great if we could give this a final review and merge it this week. I'll talk about the new IPC cache mechanism in my conference call talk today, and while the users can check my slides and the man pages it would be good to have the relevant flags on the olcf docs as well.

I think the only thing I'm on the fence about is whether the broad "GPU Aware MPI" section should recommend users set `GTL_DISABLE_HSA_IPC_SIGNAL_CACHE=0` for 8.1.33 and 9.0.1 or not. I think we should definitely recommend it for 9.1.0. We have a lot more testing of it now and I'm confident in it for that version. I don't have as much confidence in the initial implementation  in 8.1.33/9.0.1

systems/frontier_user_guide.rst
```diff
@@ -4015,6 +4019,21 @@ If it is necessary to have bit-wise reproducible results from these libraries, i
...
+
+Another enhancement in GPU-Aware MPI that relates to IPC, has to do with multiple messages in flight. Prior to Cray MPICH 8.1.33, if the maximum messages in flight at any given time exceeded ``MPICH_GPU_IPC_CACHE_MAX_SIZE``, the program would abort immediately. This has been resolved, and IPC cache handles are evicted as the number of messages in flight increase beyond ``MPICH_GPU_IPC_CACHE_MAX_SIZE``. However, a new issue exists where memory leaks can occur if GPU buffers are repeatedly created and destroyed while the number of messages in flight exceed ``MPICH_GPU_IPC_CACHE_MAX_SIZE``. To mitigate this, it is recommended to set ``MPICH_GPU_IPC_CACHE_MAX_SIZE`` to a value larger than the maximum anticipated messages in flight. The default value is ``50``. If setting ``MPICH_GPU_IPC_CACHE_MAX_SIZE > 50``, the following parameters should be adjusted accordingly to maintain a decreasing order of magnitude: ``GTL_MAX_SIG_HANDLER_LEN`` >  ``GTL_IPC_SIGNAL_POOL_TABLE_LEN`` >= ``MPICH_GPU_IPC_CACHE_MAX_SIZE``. For example: GTL_MAX_SIG_HANDLER_LEN=1024, GTL_IPC_SIGNAL_POOL_TABLE_LEN=50, MPICH_GPU_IPC_CACHE_MAX_SIZE=50.
+
```
trey-ornl, line 4033: Add double-back-quote formatting of the above example?

systems/frontier_user_guide.rst
```diff
@@ -4015,6 +4019,21 @@ If it is necessary to have bit-wise reproducible results from these libraries, i
 
 ------
 
+.. _gpu-ipc-tips-tricks:
+
+GPU inter-process communication (IPC) considerations
```
trey-ornl, line 4023: I would like this whole section to be more structured, more of a decision tree based on cray-mpich version. But consider this a feature request for a future revision. Let's get this info out there as is.

systems/frontier_user_guide.rst
trey-ornl: Github won't let me resolve this. @ashesh2512, can you?

ashesh2512: I did it in https://github.com/olcf/olcf-user-docs/pull/1081/changes/d939179d3fd2ef61db85b41c8da5588385089d93

But something got messed up - the commit doesn't belong to a branch apparently. Let me check.

ashesh2512: It should be resolved now. Can you check if it shows up on your end?

trey-ornl:
> I think the only thing I'm on the fence about is whether the broad "GPU Aware MPI" section should recommend users set `GTL_DISABLE_HSA_IPC_SIGNAL_CACHE=0` for 8.1.33 and 9.0.1 or not. I think we should definitely recommend it for 9.1.0. We have a lot more testing of it now and I'm confident in it for that version. I don't have as much confidence in the initial implementation in 8.1.33/9.0.1

I think this decision is the final roadblock. @abbotts, I leave this to you. I'm good with everything else.

systems/frontier_user_guide.rst
trey-ornl: The code looks good, but the Github _comment_ still appears not resolved. Call it a Github bug and move on?

systems/frontier_user_guide.rst
ashesh2512: I have removed `GTL_DIABLE_HSA_IPC_SIGNAL_CACHE` from here

michael-sandoval (approved):
After external discussion, this now is fully approved

PR #1085, opened 2026-05-29
Quantum: updated IQM references and partially updated software section

michael-sandoval:
Shifted from meetiqm.com to iqm.tech

Updated PennyLane guide for Frontier and removed deprecated recipes


jw676 (approved):
lgtm

PR #1087, opened 2026-06-17
update RCCL documentation with latest guidance and module information

abbotts:
This is a significant refactoring of the RCCL related docs to:
* point users at the upcoming rccl-net-plugin module
* update build and env with current BKC from the ORNL/HPE/AMD collaboration
* Reorganize pyTorch section to de-emphasize the manual plugin configuration

### DIFF
--- a/software/analytics/pytorch_frontier.rst
+++ b/software/analytics/pytorch_frontier.rst
@@ -593,7 +593,7 @@ In a nutshell: NVMe > Orion >> NFS.
 RCCL Network Plugin
 -------------------
 
-RCCL internally only supports inter-node communication using TCP/IP Sockets, which does not scale to the large job sizes on Frontier.
+RCCL defaults to inter-node communication using TCP/IP Sockets, which does not scale to the large job sizes on Frontier.
 In order to use the high-speed Slingshot network RCCL requires a network plugin that is dynamically loaded during RCCL initialization.
 The recommended way to use this plugin is to load the ``rccl-net-plugin`` module after having loaded ``rocm``.
 This module will configure the environment to use the plugin and will also set some recommended environment variables for the Slingshot network stack.
--- a/software/analytics/pytorch_frontier.rst
+++ b/software/analytics/pytorch_frontier.rst
@@ -266,7 +266,7 @@ Both scripts below use ``DistributedDataParallel`` and can run across multiple n
            os.environ['LOCAL_RANK'] = str(local_rank)
            os.environ['MASTER_ADDR'] = str(args.master_addr)
            os.environ['MASTER_PORT'] = str(args.master_port)
-           os.environ['NCCL_SOCKET_IFNAME'] = 'hsn0'
+           os.environ['NCCL_SOCKET_IFNAME'] = 'hsn0,hsn1,hsn2,hsn3' # if you see hangs, try using only hsn0
 
            dist.init_process_group(
                backend="nccl",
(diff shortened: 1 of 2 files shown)


abbotts:
@michael-sandoval , @hagertnl , @markstock - this is my initial attempt to integrate the newest RCCL plugin information into the documentation. This is a significant refactoring so should be reviewed by as many eyes as practical.

This shouldn't be merged and rolled out until after the `rccl-net-plugin` module is available, which might be a couple weeks still.

software/analytics/pytorch_frontier.rst
```diff
@@ -588,11 +590,48 @@ More information on how to use ``sbcast`` and ``conda-pack`` to move your enviro
 
 In a nutshell: NVMe > Orion >> NFS.
 
-AWS-OFI-RCCL Plugin
+RCCL Network Plugin
 -------------------
 
-The `AWS-OFI-RCCL plugin <https://github.com/ROCm/aws-ofi-rccl>`__ enables using libfabric as a network provider while running AMD's RCCL based applications.
-This plugin can be built and used by common ML/DL libraries like PyTorch to increase performance when running on AMD GPUs.
+RCCL internally only supports inter-node communication using TCP/IP Sockets, which does not scale to the large job sizes on Frontier.
```
ryanhankins: This isn't quite correct.

By default, RCCL provides inter-node communication using TCP/IP Sockets, which does not scale to the large job sizes on Frontier.

software/analytics/pytorch_frontier.rst
```diff
@@ -606,108 +645,96 @@ To build the plugin on Frontier (using ROCm 7.1.1 as an example):
...
-You also need to set the below environment variable in your runs to make sure that the ROCm RCCL library correctly identifies and loads the ``librccl-net.so`` library from the aws-ofi-rccl build. Otherwise the performance will be worse than expected because aws-ofi-rccl is not actually being used. 
+If you are concerned that the plugin isn't being used, you can force RCCL to exit if the plugin isn't initialized by setting ``NCCL_NET``.
+On Frontier, however, the network is not configured for single node jobs and the plugin will not initialize, so this setting is only recommended for multi-node jobs.
+If you want force the check, we recommend conditionally setting this variable based on the number of nodes in your job:
 
 .. code-block:: bash
 
-   export NCCL_NET_PLUGIN="librccl-net.so"
+    if (( $SLURM_JOB_NUM_NODES == 1 )); then
+        export NCCL_NET="Socket"
```
ryanhankins, line 693: Probably ok, but might be better to just not set anything.

markstock:
My only suggestions would be to remove the specific examples after "RCCL and NCCL are highly configurable with environment variables." (NCCL_DEBUG, NCCL_ALGO) and instead refer users to https://rocm.docs.amd.com/projects/rccl/en/develop/api-reference/env-variables.html

Possibly also include a statement to the effect of "RCCL's default settings and internal tuner will likely select the best protocol, algorithm, and number of channels for your collectives. The environment variables most likely to improve performance are included in the rccl-net-plugin module and also listed below."

michael-sandoval:
Since I don't want this to accidentally get merged, converting to draft

abbotts:
> My only suggestions would be to remove the specific examples after "RCCL and NCCL are highly configurable with environment variables." (NCCL_DEBUG, NCCL_ALGO) and instead refer users to https://rocm.docs.amd.com/projects/rccl/en/develop/api-reference/env-variables.html
> 
> Possibly also include a statement to the effect of "RCCL's default settings and internal tuner will likely select the best protocol, algorithm, and number of channels for your collectives. The environment variables most likely to improve performance are included in the rccl-net-plugin module and also listed below."

I tweaked the language to integrate this suggestion.

software/analytics/pytorch_frontier.rst
abbotts: I changed the language here from "only" to "RCCL defaults to inter-node..."

software/analytics/pytorch_frontier.rst
abbotts: I'm on the fence on this once. I think for clarity with the users I want to leave it how it is, because Sockets _does_ get initialized in the single node case, so I don't want users to turn on debug info, see sockets, and get confused.

If other folks have different opinions I'm willing to be outvoted, though.

abbotts:
Taking out of draft as the rccl-net-plugin module is now live on Frontier

michael-sandoval:
Doing final check but at face value looks good

michael-sandoval (approved):
LGTM and HPE approves

PR #1113, opened 2026-09-04
Update openmp compilation instructions for Frontier

t-ramz:


michael-sandoval (approved):
LGTM and double checked with Nick H.

